PR #12, opened 2020-10-11
Semantic conventions tests

justinfoote:
This PR adds unit tests to the semantic convention markdown/code generator. 

I was about to start working on refactoring this tool to add support for metrics semantic conventions, but I didn't want to get too far without at least a few unit tests, and I decided not to add all the unit tests in the same PR as the upcoming metrics work. 
So this PR should just be adding tests without changing any logic. 

There are two new directories of note:
* `semantic-conventions/src/tests`
  This is the root directory for all of the tests. They're contained in the `semconv` subdirectory, and their package structure mirrors the library itself.
* `semantic-conventions/src/tests/data`
  This directory contains test data files. Several yaml files, and one markdown. They are loaded into the tests using a pytest fixture.

The tests can be executed using the `pytest` command, like this:
```
[~/projects/build-tools/semantic-conventions] $ pytest
=============================================================== test session starts ===============================================================
platform linux -- Python 3.8.5, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: /home/justin/projects/build-tools/semantic-conventions
collected 22 items                                                                                                                                

src/tests/semconv/model/test_semantic_attribute.py ..                                                                                       [  9%]
src/tests/semconv/model/test_semantic_convention.py .                                                                                       [ 13%]
src/tests/semconv/model/test_utils.py ..................                                                                                    [ 95%]
src/tests/semconv/templating/test_markdown.py .                                                                                             [100%]

=============================================================== 22 passed in 0.15s ================================================================
[~/projects/build-tools/semantic-conventions] $ 

```

I've verified that the tests run on my machine, that the docker image still builds, and that the resulting docker image still generates the same markdown as before.  

I have not configured this repo with CI to execute these tests. It would be a good idea, and I'd love to address that in followup work.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/utils.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/utils.py
@@ -33,7 +33,7 @@ def validate_id(semconv_id, position):
 def validate_values(yaml, keys, mandatory=()):
     """ This method checks only valid keywords and value types are used
     """
-    unwanted = list(set(yaml) - set(keys))
+    unwanted = [k for k in yaml.keys() if k not in keys]
     if unwanted:
         position = yaml.lc.data[unwanted[0]]
         msg = "Invalid keys: {}".format(unwanted)
--- a/semantic-conventions/requirements.txt
+++ b/semantic-conventions/requirements.txt
@@ -7,3 +7,6 @@ ruamel.yaml.clib==0.2.0
 typed-ast==1.4.1
 typing-extensions==3.7.4.2
 Jinja2==2.11.2
+
+pytest==6.1.1
+ipdb==0.13.4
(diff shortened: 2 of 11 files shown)


semantic-conventions/src/opentelemetry/semconv/model/utils.py
```diff
@@ -33,7 +33,7 @@ def validate_id(semconv_id, position):
 def validate_values(yaml, keys, mandatory=()):
     """ This method checks only valid keywords and value types are used
     """
-    unwanted = list(set(yaml) - set(keys))
+    unwanted = [k for k in yaml.keys() if k not in keys]
```
justinfoote, line 36: Note: this is a slight change to existing logic, but it is only used for formatting an exception, and the change I made is only to ensure a consistently-ordered message, so it can be safely asserted.

thisthat (approved):
Thank you for showing the tool some love :raised_hands:
I will also follow up this PR by porting more tests so you can more safely introduce the metric support :)

arminru (approved):

PR #21, opened 2020-11-05
Add unit semantic convention generator

justinfoote:
This PR adds a new semantic convention type: units. 
There will only be the one table of units in the spec, but because we're expecting every language to create common constants for metric instrument units (see: [#1177](https://github.com/open-telemetry/opentelemetry-specification/pull/1177), I'd like to make it as easy as possible to generate the constants in code.

The unit yaml looks something like this:
```yaml
groups:
  - id: units
    type: units 
    brief: Specification of commonly used units.
    members:
    - id: percent
      brief: fraction of a total
      value: "%"
    - id: nanosecond
      brief: time
      value: NS
    - id: connections
      brief: connections
      value: "{connections}"
```

And the resulting markdown looks like this:

| Name        | Kind of Quantity         | Unit String   |
| ------------| ----------------         | -----------   |
| percent | fraction of a total | `%` |
| nanosecond | time | `NS` |
| connections | connections | `{connections}` |

There are a few concepts to check out:
* I've created four new classes to define out (currently four) semantic convention types (span, resource, metric, unit). These classes share yaml validation logic, but they each define their own allowed/required keys. The SemanticConvention base class acts as a factory for the four subclasses.
* I removed the `@dataclass` annotation from SemanticConvention, and added a custom constructor that takes a single yaml node as an argument. 

I've added tests for the new functionality, and I've manually run the markdown generator against the opentelemetry-specification repo, and it doesn't have any side-effects.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -62,10 +62,13 @@ def parse_semantic_convention_type(type_value):
     if type_value is None:
         return SpanSemanticConvention
     enum_map = {
-        "span": SpanSemanticConvention,
-        "resource": ResourceSemanticConvention,
-        "metric": MetricSemanticConvention,
-        "units": UnitSemanticConvention,
+        cls.TYPE_VALUE: cls
+        for cls in [
+            SpanSemanticConvention,
+            ResourceSemanticConvention,
+            MetricSemanticConvention,
+            UnitSemanticConvention,
+        ]
     }
     return enum_map.get(type_value)
 
@@ -98,7 +101,7 @@ def SemanticConvention(group):
     # First, validate that the correct fields are available in the yaml
     convention_type.validate_keys(group)
     model = convention_type(group)
-    # Also, validate that the value of the fields is acceptable
+    # Also, va`lidate that the value of the fields is acceptable
     model.validate_values()
     return model
 
@@ -176,6 +179,8 @@ class BaseSemanticConvention(ValidatableYamlNode):
 
 
 class ResourceSemanticConvention(HasAttributes, BaseSemanticConvention):
+    TYPE_VALUE = "resource"
+
     allowed_keys = (
         "id",
         "type",
@@ -193,6 +198,8 @@ class ResourceSemanticConvention(HasAttributes, BaseSemanticConvention):
 
 
 class SpanSemanticConvention(HasAttributes, BaseSemanticConvention):
+    TYPE_VALUE = "span"
+
     allowed_keys = (
         "id",
         "type",
@@ -216,6 +223,8 @@ class SpanSemanticConvention(HasAttributes, BaseSemanticConvention):
 
 
 class UnitSemanticConvention(BaseSemanticConvention):
+    TYPE_VALUE = "units"
+
     allowed_keys = (
         "id",
         "type",
@@ -229,6 +238,8 @@ class UnitSemanticConvention(BaseSemanticConvention):
 
 
 class MetricSemanticConvention(BaseSemanticConvention):
+    TYPE_VALUE = "metric"
+
     allowed_keys = []
 
 
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -49,21 +49,17 @@ class SpanKind(Enum):
         return kind_map.get(span_kind_value)
 
 
-class SemanticConventionType(Enum):
-    SPAN = 1
-    RESOURCE = 2
-    METRIC = 3
-
-    @staticmethod
-    def parse(type_value):
-        # Gracefully transition to the new types
-        if type_value is None:
-            return SemanticConventionType.SPAN
-        enum_map = {
-            "span": SemanticConventionType.SPAN,
-            "resource": SemanticConventionType.RESOURCE,
-        }
-        return enum_map.get(type_value)
+def parse_semantic_convention_type(type_value):
+    # Gracefully transition to the new types
+    if type_value is None:
+        return SpanSemanticConvention
+    enum_map = {
+        "span": SpanSemanticConvention,
+        "resource": ResourceSemanticConvention,
+        "metric": MetricSemanticConvention,
+        "units": UnitSemanticConvention,
+    }
+    return enum_map.get(type_value)
 
 
 @dataclass
@@ -71,16 +67,19 @@ class SemanticConvention:
     """ Contains the model extracted from a yaml file
     """
 
-    semconv_id: str
-    type: SemanticConventionType
-    brief: str
-    note: str
-    prefix: str
-    extends: str
-    span_kind: SpanKind
-    attrs_by_name: "Dict[str, SemanticAttribute]"
-    constraints: "Set[Union[Include, AnyOf]]"
-    _position: List[int]
+    mandatory_keys = ("id", "brief")
+
+    def __init__(self, group):
+        self.semconv_id = group["id"].strip()
+        # self.type = group['type']
+        self.brief = str(group["brief"]).strip()
+        self.note = group.get("note", "").strip()
+        self.prefix = group.get("prefix", "")
+        self.extends = group.get("extends", "").strip()
+        self.constraints = SemanticConvention.parse_constraint(
+            group.get("constraints", ())
+        )
+        self._position = [group.lc.line, group.lc.col]
 
     @property
     def attributes(self):
@@ -90,23 +89,9 @@ class SemanticConvention:
     def parse(yaml_file):
         yaml = YAML().load(yaml_file)
         models = []
-        available_keys = (
-            "id",
-            "type",
-            "brief",
-            "note",
-            "prefix",
-            "extends",
-            "span_kind",
-            "attributes",
-            "constraints",
-        )
-        mandatory_keys = ("id", "brief")
         for group in yaml["groups"]:
-            validate_values(group, available_keys, mandatory_keys)
-            validate_id(group["id"], group.lc.data["id"])
-            type = group.get("type")
-            if type is None:
+            type_value = group.get("type")
+            if type_value is None:
                 line = group.lc.data["id"][1] + 1
                 print(
                     "Using default SPAN type for semantic convention '{}' in {} @ line {}\n".format(
@@ -114,7 +99,7 @@ class SemanticConvention:
                     ),
                     file=sys.stderr,
                 )
-            convention_type = SemanticConventionType.parse(type)
+            convention_type = parse_semantic_convention_type(type_value)
             if convention_type is None:
                 position = (
                     group.lc.data["type"] if "type" in group else group.lc.data["id"]
@@ -131,31 +116,14 @@ class SemanticConvention:
             prefix = group.get("prefix", "")
             if prefix != "":
                 validate_id(prefix, group.lc.data["prefix"])
-            position = group.lc.data["id"]
-            # Fields validation based on the type
-            if convention_type == SemanticConventionType.RESOURCE:
-                if span_kind != SpanKind.EMPTY:
-                    position = group.lc.data["span_kind"]
-                    msg = "Resources cannot have span_kind: {}".format(
-                        group.get("span_kind")
-                    )
-                    raise ValidationError.from_yaml_pos(position, msg)
-            model = SemanticConvention(
-                semconv_id=group["id"].strip(),
-                type=convention_type,
-                brief=str(group["brief"]).strip(),
-                note=group.get("note", "").strip(),
-                prefix=prefix.strip(),
-                extends=group.get("extends", "").strip(),
-                span_kind=span_kind,
-                attrs_by_name=SemanticAttribute.parse(prefix, group.get("attributes"))
-                if "attributes" in group
-                else {},
-                constraints=SemanticConvention.parse_constraint(
-                    group.get("constraints", ())
-                ),
-                _position=position,
+
+            # First, validate that the correct fields are available in the yaml
+            validate_values(
+                group, convention_type.allowed_keys, convention_type.mandatory_keys
             )
+            model = convention_type(group)
+            # Also, validate that the value of the fields is acceptable
+            model.validate()
             models.append(model)
         return models
 
@@ -223,6 +191,14 @@ class SemanticConvention:
             [attr for attr in self.attributes if attr.required == Required.CONDITIONAL]
         )
 
+    def validate(self):
+        """
+        Subclasses may provide additional validation. 
+        This method should raise an exception with a descriptive
+        message if the semantic convention is not valid.
+        """
+        validate_id(self.semconv_id, self._position)
+
     @staticmethod
     def unique_attr(l: list) -> list:
         output = []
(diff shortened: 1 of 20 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -131,31 +121,12 @@ def parse(yaml_file):
...
-                    group.get("constraints", ())
-                ),
-                _position=position,
-            )
+
+            # First, validate that the correct fields are available in the yaml
+            convention_type.validate_keys(group)
+            model = convention_type(group)
+            # Also, validate that the value of the fields is acceptable
+            model.validate_values()
```
justinfoote, line 129: These lines appear to have gotten much simpler, but the complexity is moved from here to the constructor for the class that's returned from the `parse_semantic_convention_type` function. 

This is because the required parameters for each type of convention is different, and passing a single yaml node to the constructor allows the subclasses to own their unique implementations.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -376,14 +401,15 @@ def _populate_extends_single(
...
                 parent_attributes[ext_attr.fqn] = ext_attr.inherit_attribute()
             # By induction, parent semconv is already correctly sorted
             parent_attributes.update(
                 SemanticConventionSet._sort_attributes_dict(semconv.attrs_by_name)
             )
-            semconv.attrs_by_name = parent_attributes
-        else:  # No parent, sort of current attributes
+            if parent_attributes or semconv.attributes:
+                semconv.attrs_by_name = parent_attributes
+        elif semconv.attributes:  # No parent, sort of current attributes
```
justinfoote, line 412: This is added so we only try to sort attributes for semantic conventions that actually have attributes (spans and resources).

semantic-conventions/src/opentelemetry/semconv/model/utils.py
```diff
@@ -48,3 +48,34 @@ def check_no_missing_keys(yaml, mandatory):
         position = yaml.lc.data[list(yaml)[0]]
         msg = "Missing keys: {}".format(missing)
         raise ValidationError.from_yaml_pos(position, msg)
+
+
+class ValidatableYamlNode:
```
justinfoote, line 53: This is the base class for SemanticConvention (and also UnitMember). In the future, I expect it to be the base class for MetricLabel. 
It owns validation of keys (that all the mandatory keys are present in a yaml node, and that only allowed keys are present), and the validation of values (that the values of the constructed object are allowed).

It may be worth making SemanticAttribute derive from this class, but I didn't do that work now.

semantic-conventions/src/tests/semconv/templating/test_markdown.py
```diff
@@ -36,7 +35,8 @@ def testRef(self):
         renderer._render_single_file(content, md, output)
         with open(self.load_file("markdown/ref/expected.md"), "r") as markdown:
             expected = markdown.read()
-        self.assertEqualWithDiff(expected, output.getvalue())
+
+        assert output.getvalue() == expected
```
justinfoote, line 39: I _almost_ refactored this whole class, but I settled for just using pytest style assertions, which give much more usable messages (in my opinion). Luckily, it plays well with unittest. :)

justinfoote (commented):
@thisthat - I would especially your feedback on this PR.  I see that you and I don't have similar styles, but I tried to match the style of the project at least a little. 
I have some future changes in mind for the metrics work.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -49,72 +58,53 @@ def parse(span_kind_value):
...
-    span_kind: SpanKind
-    attrs_by_name: "Dict[str, SemanticAttribute]"
-    constraints: "Set[Union[Include, AnyOf]]"
-    _position: List[int]
+    def __init__(self, group):
+        super(SemanticConvention, self).__init__(group)
+
+        self.semconv_id = self.id
+        self.note = group.get("note", "").strip()
+        self.prefix = group.get("prefix", "")
```
thisthat, line 83: Although it is improbable that users will use the `:>` notation, I would still keep the `stip` to avoid newlines in the prefix.
```suggestion
        self.prefix = group.get("prefix", "").strip()
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -49,72 +58,53 @@ def parse(span_kind_value):
...
+        self.extends = group.get("extends", "").strip()
+        self.constraints = SemanticConvention.parse_constraint(
+            group.get("constraints", ())
+        )
 
     @property
     def attributes(self):
-        return list(self.attrs_by_name.values())
+        return []
```
thisthat, line 91: This should never be called since we expect subclasses to overwrite this method, right?
If so, I would throw an exception to indicate that something fishy is happening.

thisthat (approved):
Thank you very much for your great contribution! :rocket: 
I noticed that we have different styles but I believe the more we work on this tool, the faster we will converge to a common one :)

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
Oberon00: If you intend to go this way, consider using https://docs.python.org/3/library/abc.html#abc.abstractmethod and the abc module.

justinfoote: This method is used downstream in the SemanticConventionSet to sort attributes. That refactor felt a little too big and rabbit-holey for this PR, and there's a proposal ([#1113](https://github.com/open-telemetry/opentelemetry-specification/issues/1113)) in flight to standardize on "attributes" for metrics as well, so I thought we mightend up undoing the refactor anyway in the very near future.

semantic-conventions/src/opentelemetry/semconv/model/utils.py
```diff
@@ -48,3 +48,34 @@ def check_no_missing_keys(yaml, mandatory):
...
+
+    allowed_keys = []
+    mandatory_keys = ["id", "brief"]
```
aabmass, line 56: nit consider immutable tuples

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -49,72 +58,53 @@ def parse(span_kind_value):
...
-    brief: str
-    note: str
-    prefix: str
-    extends: str
-    span_kind: SpanKind
-    attrs_by_name: "Dict[str, SemanticAttribute]"
-    constraints: "Set[Union[Include, AnyOf]]"
-    _position: List[int]
+    def __init__(self, group):
+        super(SemanticConvention, self).__init__(group)
```
aabmass, line 79:
```suggestion
        super().__init__(group)
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -248,6 +219,60 @@ def has_attribute_constraint(self, attr):
...
+class UnitSemanticConvention(SemanticConvention):
+    allowed_keys = [
+        "id",
+        "type",
+        "brief",
+        "members",
+    ]
+
+    def __init__(self, group):
+        super(UnitSemanticConvention, self).__init__(group)
```
aabmass, line 268:
```suggestion
        super().__init__(group)
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -131,31 +121,12 @@ def parse(yaml_file):
...
-                else {},
-                constraints=SemanticConvention.parse_constraint(
-                    group.get("constraints", ())
-                ),
-                _position=position,
-            )
+
+            # First, validate that the correct fields are available in the yaml
+            convention_type.validate_keys(group)
+            model = convention_type(group)
```
aabmass, line 127: Should the convention class validate itself in the constructor so you don't have to remember to validate separately?

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -248,6 +219,60 @@ def has_attribute_constraint(self, attr):
...
+        "brief",
+        "note",
+        "prefix",
+        "extends",
+        "attributes",
+        "constraints",
+    ]
+
+    def __init__(self, group):
+        super(ResourceSemanticConvention, self).__init__(group)
```
aabmass, line 236:
```suggestion
        super().__init__(group)
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -248,6 +219,60 @@ def has_attribute_constraint(self, attr):
...
+        "note",
+        "prefix",
+        "extends",
+        "span_kind",
+        "attributes",
+        "constraints",
+    ]
+
+    def __init__(self, group):
+        super(SpanSemanticConvention, self).__init__(group)
```
aabmass, line 254:
```suggestion
        super().__init__(group)
```

semantic-conventions/src/opentelemetry/semconv/model/unit_member.py
```diff
@@ -0,0 +1,22 @@
+from opentelemetry.semconv.model.utils import ValidatableYamlNode
+
+
+class UnitMember(ValidatableYamlNode):
+
+    allowed_keys = ["id", "brief", "value"]
+    mandatory_keys = allowed_keys
+
+    def __init__(self, node):
+        super(UnitMember, self).__init__(node)
```
aabmass, line 10:
```suggestion
        super().__init__(node)
```

aabmass (approved):
💯

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
justinfoote: I considered this. The logical place to put this would be in `ValidatableYamlNode.__init__`, because that superclass owns the validation logic. **..._but_**, the subclasses are likely adding attributes in their own init methods, so the correct place to call it would be _after_ the superclass and subclass init methods had completed.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -49,72 +58,53 @@ def parse(span_kind_value):
         return kind_map.get(span_kind_value)
 
 
-class SemanticConventionType(Enum):
-    SPAN = 1
-    RESOURCE = 2
-    METRIC = 3
```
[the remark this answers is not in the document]
justinfoote: I prefer to have this map return something that can be used (like the correct class for the semantic convention). I don't see the value in having the bare strings mapped to integers.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
justinfoote: :exploding_head: This is amazing! I haven't used python professionally since python 2.7 was the thing, and everyone was on the fence about migrating to python 3, so I'm out of the loop on all the new fancy features.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
aabmass: maybe in a factory then?

justinfoote: I moved the logic to a factory function.
This triggered some renaming (so I wouldn't have a naming collision between `def SemanticConvention` and `class SemanticConvention`), which had some cascading effects.  
I removed several `import SemanticConvention` lines where we were only using the imported class for type declarations, and I refactored references to `SemanticConvention.parse` to a new `parse_semantic_convention_groups` function (and in the process I split the parsing concern away from the data struct concern).

Oberon00 (changes requested):
Please don't remove the enum. The integer values don't matter, but the enum is safer against typos. Even if we don't use static checking like pylint or mypy (yet), we at least would get a runtime error if we type `SemanticConventionType.SAN` whereas with `== "san"` we would just get `False`.

Also the enum is a nice documentation of the possible types.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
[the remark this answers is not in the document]
Oberon00: Posted my comment about this in the wrong place, sorry: https://github.com/open-telemetry/build-tools/pull/21#pullrequestreview-526042782

> Please don't remove the `SemanticConventionType` enum. The integer values don't matter, but the enum is safer against typos. Even if we don't use static checking like pylint or mypy (yet), we at least would get a runtime error if we type `SemanticConventionType.SAN` whereas with `== "san"` we would just get `False`.

> Also the enum is a nice documentation of the possible types.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -98,7 +101,7 @@ def SemanticConvention(group):
     # First, validate that the correct fields are available in the yaml
     convention_type.validate_keys(group)
     model = convention_type(group)
-    # Also, validate that the value of the fields is acceptable
+    # Also, va`lidate that the value of the fields is acceptable
```
Oberon00, line 104: Typo

```suggestion
    # Also, validate that the value of the fields is acceptable
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -176,6 +179,8 @@ def validate_values(self):
 
 
 class ResourceSemanticConvention(HasAttributes, BaseSemanticConvention):
+    TYPE_VALUE = "resource"
```
Oberon00, line 182: The name TYPE_VALUE is unclear to me, maybe GROUP_TYPE_NAME (a group of this type has this name). Although that also sounds confusing

aabmass (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
justinfoote: :+1:

Oberon00:
I think this PR is too huge to properly review it. It seems to be almost a complete rewrite. If others are fine with still merging this, I won't object, personally I would be much happier with smaller pieces if possible (e.g. only have the base class refactoring in a PR, without the addition of Unit, command line interface changes, etc).

But I think we should decide on this PR soon, as having a huge refactoring looming might discourage any other changes, as they would lead to merge conflicts.

jmacd (approved):

jmacd:
I can't really review this either, but would be happy to see it merged presuming the output is good and others who have looked at it agree.

arminru:
Same as @jmacd.

Both @aabmass and @thisthat already reviewed and approved it - thanks for that! If no one objects by tomorrow, let's merge it.

cc @open-telemetry/specs-approvers

justinfoote:
@open-telemetry/specs-approvers As @arminru suggested above, let's go ahead and merge.

arminru:
Thanks for your contribution and for sticking it out, @justinfoote!

PR #22, opened 2020-11-05
Add CI workflow to run python tests in semantic convention generator

justinfoote:
This simple PR just adds a new github actions workflow to run the python tests when a PR is made against the semantic convention directory.  It also fixes a typo in the name of the existing workflow for semantic conventions.


thisthat (approved):

.github/workflows/semconvgen-tests.yml
```diff
@@ -0,0 +1,23 @@
+name: Semantic Convention Generator Run Tests
+on:
+  push:
+    tags: [ '**' ]
```
Oberon00, line 4: Do we want to run this on tags? If so, is there a way to skip publishing if the test workflow fails?

.github/workflows/semconvgen-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v2
+    - name: install dependencies
+      run: |
+        pip install -U pip
```
Oberon00, line 20:
```suggestion
        pip install -U pip setuptools wheel
```

Oberon00 (approved):

.github/workflows/semconvgen-tests.yml
justinfoote: I thought maybe that `jobs.{job_id}.needs` was the solution here, but I guess not -- the build job just didn't run at all. I have one more thing to try.

justinfoote: That did the trick. The two jobs run in sequence now, which may cause the builds to take longer than they otherwise would, but this seems like a reasonable trade off in a project that doesn't get built that often.

.github/workflows/semconvgen-tests.yml
justinfoote: Done in d087986

.github/workflows/semconvgen.yml
```diff
@@ -1,17 +1,30 @@
...
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v2
+    - name: install dependencies
+      run: |
+        pip install -U pip setuptools wheel
+        pip install -r semantic-conventions/requirements.txt
+    - name: run tests
+      run: pytest -v
+
   build:
```
arminru, line 25:
```suggestion
  build-and-publish-docker:
```

arminru (approved):

arminru:
https://github.com/open-telemetry/build-tools/runs/1391937425

Looks good - thanks, Justin!

PR #24, opened 2020-12-15
Better markdown generation

### LINKED ISSUE #7: SemConv MD generator should not prepend "Conditional" to requirement conditions
From https://github.com/open-telemetry/opentelemetry-specification/pull/925#discussion_r484440799:

`"Required: Conditional No, but recommended"` looks odd and might be misleading.
The MD generator should only print the condition itself without prepending the word _Conditional_ in the table, unless it references a footnote.

### LINKED ISSUE #8: SemConv MD generator should write "or" between examples
From https://github.com/open-telemetry/opentelemetry-specification/pull/928#discussion_r485198089

Having an example per row makes it harder to understand that they are multiple examples and not a single one. The MD generator should use "or" instead of "<br>" to separate examples.

### LINKED ISSUE #9: SemConv MD generator should not use "enum" in attribute types
From https://github.com/open-telemetry/opentelemetry-specification/pull/926#discussion_r492054452

Attribute types that have a finite set of options, e.g. [network.transport](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes), should not report their type with the `enum` suffix.

### LINKED ISSUE #23: Generator should print examples between `[` `]` for array types

thisthat:
This PR addresses:
- Print "Conditional" only to footnotes - closes #7
- Write "or" between examples - closes #8 
- Don't print "enum" in attribute type - closes #9
- Print examples for arrays between `[]` - closes #23


thisthat:
@justinfoote FYI

Oberon00:
Can you reformat the PR description so that it auto-closes the issues? E.g. "Closes #7: Print..."

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
```diff
@@ -4,12 +4,12 @@
 <!-- semconv http -->
 | Attribute  | Type | Description  | Example  | Required |
 |---|---|---|---|---|
-| `http.method` | string | HTTP request method. | `GET`<br>`POST`<br>`HEAD` | Yes |
+| `http.method` | string | HTTP request method. | `GET` or `POST` or `HEAD` | Yes |
```
Oberon00, line 7:
```suggestion
| `http.method` | string | HTTP request method. | `GET`, `POST`, `HEAD` | Yes |
```

I would prefer commas. Using multiple `or`s sounds wrong.

Oberon00 (approved):

Oberon00:
Thank you for taking care of these long-standing issues!

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
thisthat: On second thought, I agree. Reading or multiple times sounds weird.
Maybe `or` only for two options and `,` for multiple?

Oberon00: If there is a concern that commas might make people think of arrays, I suggest: Use `or` only between the last two items of the list (like you normally do in English); or use semicolons instead of commas as separators.

semantic-conventions/src/tests/data/markdown/extend_constraint/expected.md
```diff
@@ -38,11 +38,11 @@ Some database systems may allow a connection to switch to a different `db.user`,
 |---|---|---|---|---|
 | `db.system` | string | An identifier for the database management system (DBMS) product being used. See below for a list of well-known identifiers. | `other_sql` | Yes |
 | `db.connection_string` | string | The connection string used to connect to the database. [1] | `Server=(localdb)\v11.0;Integrated Security=true;` | No |
-| `db.user` | string | Username for accessing the database. | `readonly_user`<br>`reporting_user` | No |
-| `net.peer.ip` | string | Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6) | `127.0.0.1` | Conditional<br>See below. |
```
justinfoote, line 42: 👏

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
justinfoote: I think putting a smart join in here (that uses `or` between two things, commas between more than two, and an additional `or` between the final two) seems like too much for this tool.  And I'd really prefer not to start a conversation about whether to use an oxford comma. 

I vote for semicolons. 😄

justinfoote (commented):
🎉 This is a great addition! Replacing those `<br>`s makes these tables much more clear.

I'll approve once we have the comma/semicolon/or decision made.

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
thisthat: @open-telemetry/specs-approvers what would you suggest to use?

arminru: The table header should say _Example**s**_ instead of _Example_ to make clear those are multiple, separate examples (still fine if there is only one example given).
I'd say commas and one _or_ (rather without oxford comma), just commas or just semicolons are all fine. When it says example**s** on top and also mentions the type, even commas should not confuse readers. Multiple *or*s really look odd, though.

Oberon00: Let's use semicolons!

thisthat: Done in 8053f628e25e7486766adb8a837d19ecb160a5cb

arminru (approved):
Thank you, Giovanni!

justinfoote (approved):

PR #25, opened 2021-01-21
Upgrade grpc/protoc and all deps in the protobuf builder docker

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

### DIFF
--- a/protobuf/Dockerfile
+++ b/protobuf/Dockerfile
@@ -1,46 +1,59 @@
 ARG ALPINE_VERSION=3.12
-ARG GO_VERSION=1.14.6
-ARG GRPC_GATEWAY_VERSION=1.14.6
-ARG GRPC_JAVA_VERSION=1.30.2
+ARG GO_VERSION=1.15.6
 # gRPC core version that applies to C++, C#, Objective-C, PhP, Python, Ruby
-ARG GRPC_VERSION=1.30.2
-ARG PROTOC_GEN_GO_VERSION=1.4.2
-# v1.3.1, using the version directly does not work: "tar: invalid magic"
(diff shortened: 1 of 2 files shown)


protobuf/Dockerfile
```diff
@@ -1,46 +1,59 @@
...
+        -GNinja \
+        -DBUILD_SHARED_LIBS=ON \
+        -DCMAKE_INSTALL_PREFIX=/usr \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DgRPC_INSTALL=ON \
+        -DgRPC_BUILD_TESTS=OFF \
+        ../.. && \
+    cmake --build . --target plugins && \
+    cmake --build . --target install && \
+    DESTDIR=/out cmake --build . --target install
```
tigrannajaryan, line 37: Just curious: are these changes dictated by the new version of grpc? Did they move from make to cmake?

tigrannajaryan (approved):
I don't have the tools installed locally so don't have a way to test the change. I trust you verified it, so approving.

protobuf/Dockerfile
bogdandrutu: Yes they now only use cmake and bazel for building.

bogdandrutu:
@tigrannajaryan yes verified with the proto, also added support for php, ruby, cpp, csharp and everything. Soon you will see a PR there

jpkrohling:
Did you rebase this based on the origin that we used for the Jaeger code?

bogdandrutu:
@jpkrohling yes I used the original repo that you also used as inspiration

PR #30, opened 2021-02-23
Add int and double types. Small bugfixes.

### LINKED ISSUE #13: SemConv generator should differentiate int and double type in YAML model and generated files
From https://github.com/open-telemetry/opentelemetry-specification/issues/1096:

YAML model for attributes seems to only have number type. But we have int or double as possible attribute types.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/common/common.md#attributes
https://github.com/open-telemetry/opentelemetry-specification/blob/master/semantic_conventions/syntax.md

thisthat:
This PR fixes #13 

FYI @anuraaga :)

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -49,7 +49,7 @@ class HasAttributes:
 
 def unique_attributes(attributes):
     output = []
-    for x in l:
+    for x in attributes:
         if x.fqn not in [attr.fqn for attr in output]:
             output.append(x)
     return output
@@ -237,7 +237,14 @@ class SemanticAttribute:
             and examples is not None
         ):
             examples = [examples]
-        if is_simple_type and attr_type not in ["boolean", "boolean[]"]:
+        if is_simple_type and attr_type not in [
+            "boolean",
+            "boolean[]",
+            "int",
+            "int[]",
+            "double",
+            "double[]",
+        ]:
             if examples is None or (len(examples) == 0):
                 position = attribute.lc.data[list(attribute)[0]]
                 msg = "Empty examples for {} are not allowed".format(attr_type)
@@ -275,8 +282,10 @@ class AttributeType:
         return attr_type in (
             "string",
             "string[]",
-            "number",
-            "number[]",
+            "int",
+            "int[]",
+            "double",
+            "double[]",
             "boolean",
             "boolean[]",
         )
@@ -284,8 +293,10 @@ class AttributeType:
     @staticmethod
     def type_mapper(attr_type: str):
         type_mapper = {
-            "number": int,
-            "number[]": int,
+            "int": int,
+            "int[]": int,
+            "double": float,
+            "double[]": float,
             "string": str,
             "string[]": str,
             "boolean": bool,
--- a/semantic-conventions/src/opentelemetry/semconv/templating/markdown.py
+++ b/semantic-conventions/src/opentelemetry/semconv/templating/markdown.py
@@ -121,7 +121,7 @@ class MarkdownRenderer:
             description += " [{}]".format(len(self.render_ctx.notes))
         examples = ""
         if isinstance(attribute.attr_type, EnumAttributeType):
-            if attribute.is_local and not attribute.ref:
+            if self.render_ctx.is_full or (attribute.is_local and not attribute.ref):
                 self.render_ctx.add_enum(attribute)
             example_list = attribute.examples if attribute.examples else ()
             examples = (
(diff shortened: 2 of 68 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -237,7 +237,14 @@ def parse_id(attribute):
...
             examples = [examples]
-        if is_simple_type and attr_type not in ["boolean", "boolean[]"]:
+        if is_simple_type and attr_type not in [
+            "boolean",
+            "boolean[]",
+            "int",
+            "int[]",
+            "double",
+            "double[]",
+        ]:
```
Oberon00, line 247: It seems to me that we should change this to

```suggestion
        if is_simple_type and attr_type in ["string", "string[]"]:
```

(not sure if we need the `is_simple_type and`).

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -275,17 +282,21 @@ def is_simple_type(attr_type: str):
...
+            "int",
+            "int[]",
+            "double",
+            "double[]",
             "boolean",
             "boolean[]",
         )
 
     @staticmethod
     def type_mapper(attr_type: str):
```
Oberon00, line 294: This name seems odd for a function. I want to suggest `parse_base_type`, but I don't know if the term "base type" is the right one for this `int[]` -> `int`, etc. mapping.

semantic-conventions/src/opentelemetry/semconv/templating/markdown.py
```diff
@@ -399,6 +399,12 @@ def _render_table(self, semconv, parameters, output):
                 continue
             if self.render_ctx.is_full or attr.is_local:
                 attr_to_print.append(attr)
+        if self.render_ctx.group_key is not None and not attr_to_print:
```
Oberon00, line 402: Why don't we just:

```suggestion
        if not attr_to_print:
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -275,17 +282,21 @@ def is_simple_type(attr_type: str):
...
             "string",
             "string[]",
-            "number",
-            "number[]",
+            "int",
+            "int[]",
+            "double",
+            "double[]",
             "boolean",
             "boolean[]",
```
Oberon00, line 290: If you make the dictionary in type_mapper a (static) constant, you can just do `if attr_type in that_dict` (`in` checks the keys with a dictionary right-hand side)

Oberon00 (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
thisthat: I am more in favor of explicitly excluding the types for which an example is not needed. 
`is_simple_type` is required because enums do not need an example.

semantic-conventions/src/opentelemetry/semconv/templating/markdown.py
thisthat: because there could be the case where a semantic convention extends an existing one without introducing any new attribute but only constraints. An example of this is `faas.http`.

Oberon00: Hmm, but how is that related to the group_key? Isn't that the filter key like `tech-specific` or `connection-level`?

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
Oberon00: OK for me. But maybe add a comment like: "# Don't enforce examples for numeric and boolean".

semantic-conventions/src/opentelemetry/semconv/templating/markdown.py
thisthat: yes! If you filter over a key and you don't get any result, you will receive a warning.

Oberon00: OK, I think I understand now. The problem is that `attr_to_print` being empty is fine if you also have (any-of) constraints, unless you added a filter, because then you probably meant to filter out only some but not all attributes.

Oberon00:
@open-telemetry/specs-approvers Can we merge this?

arminru (commented):
@thisthat I think this is our first breaking change with the generator. Could you please either add a changelog section in the README or a separate file for tracking these changes? Users will have to update their YAML files accordingly.

SergeyKanzhelev (approved):

PR #36, opened 2021-03-18
Fix markdown generator for int enums and add test

arminru:
I just noticed that it still says `number` here:
https://github.com/open-telemetry/opentelemetry-specification/blob/b67faa4e0106cac194a837e3d419ae4fab9fc1c4/specification/trace/semantic_conventions/rpc.md#grpc-attributes

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -360,7 +360,7 @@ class EnumAttributeType:
         return self.enum_type
 
     @staticmethod
-    def __is_valid_enum_value__(val):
+    def is_valid_enum_value(val):
         return isinstance(val, int) or isinstance(val, str)
 
     @staticmethod
@@ -394,7 +394,7 @@ class EnumAttributeType:
             mandatory_keys = ["id", "value"]
             for member in attribute_type["members"]:
                 validate_values(member, allowed_keys, mandatory_keys)
-                if not EnumAttributeType.__is_valid_enum_value__(member["value"]):
+                if not EnumAttributeType.is_valid_enum_value(member["value"]):
                     raise ValidationError(0, 0, "Invalid value used in enum: <{}>".format(member["value"]))
                 members.append(
                     EnumMember(
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -361,7 +361,7 @@ class EnumAttributeType:
 
     @staticmethod
     def is_valid_enum_value(val):
-        return isinstance(val, int) or isinstance(val, str)
+        return isinstance(val, (int, str))
 
     @staticmethod
     def parse(attribute_type):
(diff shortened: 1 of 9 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -272,7 +272,7 @@ class AttributeType:
     @staticmethod
     def get_type(t):
         if isinstance(t, numbers.Number):
```
Oberon00, line 274: Technically this could be double. Either we reject double outright (preferable IMHO), or we need to distinguish int and float here and generate int/double accordingly.

arminru: I'd reject double right away and if one would ever specify floating point numbers as enum values we could just treat them as strings.

arminru: Changed it to reject doubles in 8374bb4.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -271,8 +271,12 @@ class AttributeType:
 
     @staticmethod
     def get_type(t):
-        if isinstance(t, numbers.Number):
+        if isinstance(t, int):
             return "int"
+        if isinstance(t, numbers.Number):
+            raise ValidationError(
```
Oberon00, line 277: Shouldn't this be validated "eagerly"? I.e., even if get_type is not called on the enum, there should be an error (maybe get_type is currently always called, but logically it could make more sense to have this somewhere else).

arminru: `get_type` is called only from `EnumAttributeType.parse`, so I considered this part of enum parsing but I can move the validation directly into `EnumAttributeType.parse` to make it cleaner.

Oberon00: In that case, just renaming the function or maybe inlining it might be more clear.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -363,6 +359,10 @@ def __init__(self, custom_values, members, enum_type):
     def __str__(self):
         return self.enum_type
 
+    @staticmethod
+    def __is_valid_enum_value__(val):
```
Oberon00, line 363: Please don't use double-underscores, they are meant for certain special methods.

Just use 
```suggestion
    def is_valid_enum_value(val):
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
arminru: Just noticed it's also used in one other place, so I left `get_type` itself as such.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -359,6 +359,10 @@ def __init__(self, custom_values, members, enum_type):
     def __str__(self):
         return self.enum_type
 
+    @staticmethod
+    def is_valid_enum_value(val):
+        return isinstance(val, int) or isinstance(val, str)
```
Oberon00, line 364:
```suggestion
        return isinstance(val, (int, str))
```

Oberon00 (approved):

thisthat (approved):

PR #43, opened 2021-05-06
Update semconv tool with stability fields and MD render.

### LINKED ISSUE #35: semantic convention marker: stability
From https://github.com/open-telemetry/opentelemetry-specification/issues/1497#issuecomment-800857206 there are cases we may want to stabilize a subset of attributes in a given category. It would be nice if we can just mark attributes as stable so the generator handles this for us. This is less important for MD generation and more for code, where stable attributes will need to be rendered differently (different packages for example).

thisthat:
This PR fixes #35 introducing 3 level of stability:
- Stable
- Experimental
- Deprecated

The stability can be expressed at semantic convention and attribute level.
If an attributes doesn't have a stability marker, it inherits the stability of the semantic convention.
If no stability marker is set, it defaults to "stable".

The PR also improves the code generator adding support for Markdown rendering.
For example, given the following Yaml
```
groups:
  - id: test
    type: resource
    brief: test
    attributes:
      - id: provider
        type: string
        brief: >
           	Name *of* **the** `cloud` provider.
```
and the following template
```
{%- for attribute in attributes %}
 {{attribute.brief | render_markdown(code="{{@code {0}}}", link="<a href='{0}'>{1}</a>", emphasis="<i>{0}</i>", strong="<b>{0}</b>") }}.
{%- endfor %}
```
the code generator will produce:
```md
Name <i>of</i> <b>the</b> {@code cloud} provider.
```


arminru:
Please also make sure to update https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/syntax.md once this is merged (or in parallel).

Oberon00 (approved):
LGTM in general. The design decision that can be discussed here is to merge deprecation and stability into a single enum, so once deprecated you can't distinguish if an attribute was formerly experimental or stable.

PR #49, opened 2021-06-28
Add go semantic convention generator

Aneurysm9:
This is the driver used to generate semantic convention constants in the Go SDK.  We would like to move it here to enable sharing the implementation with the collector.  Templates will remain in their respective repositories.


weyert:
Sorry, how does this differ from the existing generator using the Python-based tool?

bogdandrutu:
> Sorry, how does this differ from the existing generator using the Python-based tool? Does this run the same tool but instead of using shell script it uses Go?

The python based generates markdowns, this tool generates golang constants for these values.

bogdandrutu:
@MrAlias can you review?

weyert:
Yes, but ain’t this the same what the Java and Javascript projects are doing by using this tool to generate source files for enums/constants based on the semantic convention yaml  files?

Aneurysm9:
This runs the existing Python-based processor and performs some post-processing to ensure that the generated constant identifiers conform to the go idiom with respect to initialisms and acronyms.

go-semconv/generator.go
```diff
@@ -0,0 +1,437 @@
...
+	)
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("unable to render template: %w", err)
+	}
+
+	err = os.MkdirAll(cfg.outputPath, 0700)
+	if err != nil {
+		return fmt.Errorf("unable to create output directory %s: %w", cfg.outputPath, err)
+	}
```
MrAlias, line 173: nit: running this prior to the "expensive" rendering operation would allow a quick failure if this fails.

go-semconv/generator.go
```diff
@@ -0,0 +1,437 @@
...
+			versions = append(versions, ver)
+		}
+	}
+
+	// Sort it according to semver rules.
+	sort.Sort(versions)
+
+	if len(versions) == 0 {
+		return "", fmt.Errorf("no version tags found in the specification repo at %s", cfg.inputPath)
+	}
```
MrAlias, line 226: nit: this should probably be done before the sort.

Oberon00 (dismissed):
> This runs the existing Python-based processor and performs some post-processing to ensure that the generated constant identifiers conform to the go idiom with respect to initialisms and acronyms.

I think this should not be necessary. The Jinja 2 language that is used in templates is very flexible, basically a full Python dialect. See for example https://jinja.palletsprojects.com/en/2.11.x/templates/#capitalize, or if you need a replacement table, there are also dictionary literals, see https://jinja.palletsprojects.com/en/2.11.x/templates/#literals. If there is something that the semantic convention generator cannot do, it should be much simpler to extend it than to write a post-processing tool.

Please review the `code` subcommand of the semantic convention generator and the [Jinja 2 template documentation](https://jinja.palletsprojects.com/en/2.11.x/templates).

EDIT:
Also, this tool looks like it could only work with a certain template that generates certain identifiers. The template is not bundled though. If you decide the custom tool is still needed, it might be better to maintain it in the go repository.

MrAlias (approved):

Aneurysm9:
> > This runs the existing Python-based processor and performs some post-processing to ensure that the generated constant identifiers conform to the go idiom with respect to initialisms and acronyms.
> 
> I think this should not be necessary. The Jinja 2 language that is used in templates is very flexible, basically a full Python dialect. See for example https://jinja.palletsprojects.com/en/2.11.x/templates/#capitalize, or if you need a replacement table, there are also dictionary literals, see https://jinja.palletsprojects.com/en/2.11.x/templates/#literals. If there is something that the semantic convention generator cannot do, it should be much simpler to extend it than to write a post-processing tool.

I don't speak Python, so it was much simpler for me to write this as a post-processor in Go.  It also allowed me to wrap the invocation of the generator in Go and not fight with trying to write a safe shell script.  I expect that this post-processing is only applicable to Go libraries and that the Go SDK or collector maintainers will be making any necessary changes going forward.

> Please review the `code` subcommand of the semantic convention generator and the [Jinja 2 template documentation](https://jinja.palletsprojects.com/en/2.11.x/templates).
> 
> EDIT:
> Also, this tool looks like it could only work with a certain template that generates certain identifiers. The template is not bundled though. If you decide the custom tool is still needed, it might be better to maintain it in the go repository.

This tool can work with any template.  No template is included as the Go SDK and the collector require different templates and will maintain those templates in their respective repositories.  We are currently working through the identification of a number of build tools that are useful to both the Go SDK and the collector with the plan to move them to this repository as a neutral shared location.  If that is going to be a problem then I think we need to have a new repository created for this purpose.

Oberon00:
Oh, I always forget that Go has two repositories that might want to share code.
Using go as a shell script replacement is fine for me, but even if the post-processor was easier to write than the template, I doubt it will be easier to maintain in the long run, adding another layer in another language to the build process.

If you really think this is the best solution for Go though, I don't think I have a right to stop you 😃 But I do ask you to add wording to the REAMDE that this is not something other languages should do but rather a hack that is only justified due to Go having multiple independent repositories using this.

go/tools/semconvgen/README.md
```diff
@@ -0,0 +1,23 @@
+# Go Semantic Convention Generator
+
+This tool is designed to generate constants in a semantic convention package for the Go API and the collector.
+It may be used by other systems, but it's primary function beyond invoking the template processor is to ensure that
+generated identifiers conform to Go's naming idiom, particularly with respect to initialisms and acronyms.
```
Oberon00, line 5: Please mention here that this primary function could also be fulfilled just with the Python tool.

go/tools/semconvgen/README.md
```diff
@@ -0,0 +1,23 @@
...
+Other users may be served just as well by using the template processor directly.
+
+## Usage
+
+```shell
+$ semconvgen -i <path to spec YAML> -t <path to template> -o <path to output>
```
Oberon00, line 11: The Python tool is invoked with `gen-semconv`, so maybe use a name with the same scheme, like `go-gen-semconv`

Aneurysm9:
@Oberon00 I have added a `CODEOWNERS` entry and a `README.md` for the `go` directory.  Hopefully this should clarify that these tools are intended for the Go SDK and Collector maintainers and not others.  

As far as executable naming, can we please not get stuck bikeshedding?  One is as good as the other so let's leave it as-is.

Oberon00:
The executable name was not a blocking comment 😃

alolita:
@jsuereth @anuraaga @carlosalberto can we please get your reviews on this PR? Ty :-)

anuraaga:
I'd suggest to keep in Go repository if it's possible for sharing with Collector.

There is some precedence for using code from opentelemetry-go in the collector

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b002feacaec1f1d0c634502bb53e6f7f2a4a8729/internal/aws/metrics/metric_calculator.go#L21

And I imagine there will be even more, especially when internal metrics are ported to use it. Can't the collector just reference semantic conventions through the Go SDK instead of running the tool again just for the collector? It seems somewhat similar to the java and java-instrumentation repos - java-instrumentation doesn't generate its own copy of semantic attributes to expose to instrumentation, instrumentation just uses the semantic attributes from the SDK.

Even if the generated code must be different (would be unfortunate), is there a significant downside from just going ahead and hosting the tool in opentelemetry-go? If we needed a tool shared among the Java repos I'm pretty sure that's the approach we would have taken.

alolita:
Hi @anuraaga - you read our mind :-) we plan to use this common generator for sharing conventions and tools with the Collector, Go SDK and contrib repos. This issue is a required for GA item in our upcoming Collector trace stable release. See https://github.com/open-telemetry/opentelemetry-collector/milestone/14 Ping me or @Aneurysm9 if you have any questions. Thanks for reviewing!

alolita:
@jsuereth @bogdandrutu can you please help review and merge these changes. We would like to get this PR merged asap to make progress on the tools needed. 

Btw @bogdandrutu any idea why are @open-telemetry/specs-approvers approving these build-tools? Can we have a different list of approvers?

Oberon00:
@Aneurysm9 https://github.com/open-telemetry/build-tools/pull/49#issuecomment-870741219

> I don't speak Python, so it was much simpler for me to write this as a post-processor in Go

If you like, and you point me to the template, I can try to come up with a way to handle the post-processing right in the template. That would also resolve my main gripe with this PR.

bogdandrutu:
@Oberon00 I see your point, but as @Aneurysm9 mentioned, we are not experts in go, so if you are willing to make a PR to update the python tool to generate go constants as we want we will be happy to use that.

bogdandrutu:
@anuraaga 

>I'd suggest to keep in Go repository if it's possible for sharing with Collector.
There is some precedence for using code from opentelemetry-go in the collector
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b002feacaec1f1d0c634502bb53e6f7f2a4a8729/internal/aws/metrics/metric_calculator.go#L21

I think that is not encouraged, and if you do that is on you we (collector maintainers) will not provide a direct API that accepts the otel attributes.

go/tools/semconvgen/README.md
bogdandrutu: As mentioned in another comment, happy to review a PR that helps us with that change :)

anuraaga:
> I think that is not encouraged, and if you do that is on you we (collector maintainers) will not provide a direct API that accepts the otel attributes.

@bogdandrutu I think in general when writing instrumentation, there are separate instrumentation libraries that directly use the SDK's semantic attributes interface. I'm just wondering what the difference is for the collector?

Oberon00:
@bogdandrutu 

> if you are willing to make a PR to update the python tool to generate go constants as we want we will be happy to use that.

This is a misunderstanding. The tool is already able to do that through the powerful templating language (Jinja2). but the right part is that I would be willing to provide that template if you give me a link to the one that is currently used.

jsuereth:
I want to call out an issue here:  Even *if* @Oberon00 is able to provide a Jinja template in python, will the Go SiG be able to maintain their codegen going forward?

I think it's ok for some duplication here *as long as* each SiG can service the needs they have for their own codegen.   To the extent we have a build-tools repository that's designed to service all SiG needs, I suggest we allow some divergence here and *suggest* to the Go SiG to learn Jinja, but we cannot force them.   I'd rather *not* put a hard dependency between Go-SiG + @Oberon00 having to make updates for their needs going forward.

go/tools/semconvgen/README.md
jsuereth: +1 to renaming this to go-gen-semconv.   We can follow similar conventions for other langaauge SiGs that want/need specialized codegen.

jsuereth (approved):
Approving from my standpoint on the following two conditions:

1. The tool command-line path clearly denotes "go", as suggested.
2. We open a bug task to investigate re-unifying with python.   If Go SiG is unable to find an owner who wants to learn python, I don't think we should force this, but at least we can make an attempt.   It's likely the Jinja tool is far more powerful and could lead to less maintenance overhead, but I think that's a decision for the primary maintainer/user of this tooling.

Oberon00:
> Even if @Oberon00 is able to provide a Jinja template in python, will the Go SiG be able to maintain their codegen going forward?

I think so. Jinja is not that complex and has good documentation. Even if I don't provide the adapted template, the non-adapted template still has to be maintained. Currently maintaining the template seems to mean writing a post-processor in go for it. This could still be done at a later stage if the template proves unmaintainable (which I doubt).

Oberon00 (changes requested):
@jsuereth 
> Approving from my standpoint on the following two conditions:
> * The tool command-line path clearly denotes "go", as suggested.

I agree with that condition and will block the PR until that is resolved, just to avoid this getting lost.

Aneurysm9:
> @jsuereth
> 
> > Approving from my standpoint on the following two conditions:
> > 
> > * The tool command-line path clearly denotes "go", as suggested.
> 
> I agree with that condition and will block the PR until that is resolved, just to avoid this getting lost.

The import path (`github.com/open-telemetry/build-tools/go/tools/semconvgen`) does include `go`.  The name of the executable built from this import path is irrelevant and can be whatever the end user wants.

Aneurysm9:
> > Even if @Oberon00 is able to provide a Jinja template in python, will the Go SiG be able to maintain their codegen going forward?
> 
> I think so. Jinja is not that complex and has good documentation. Even if I don't provide the adapted template, the non-adapted template still has to be maintained. Currently maintaining the template seems to mean writing a post-processor in go for it. This could still be done at a later stage if the template proves unmaintainable (which I doubt).

There is going to be a Go-specific post-processing step no matter what we do as we need to run `gofmt` against the generated code.  Ignoring that for the moment though, the work this tool performs (in terms of non-gotmt post-processing) is commonly applicable against multiple different templates.  It should not be built into each template.  Perhaps it could be built into the code that invokes Jinja as a filter function, but then we force everyone to carry those functions that are only useful to Go-based consumers and also force us to go through the Python library to update the post-processing.

If this repository is not going to be allowed to contain build tools in and for the broad range of languages we support then it needs to be renamed.  I'd rather not have to block every PR to this repo that doesn't include renaming `semantic-conventions` to `python-semconvgen`.

alolita:
@Oberon00 I'm unclear about where this discussion is going. Are you recommending all tooling for various languages supported in OpenTelemetry be developed in Python?  I don't see the need to have a single-language approach for all tooling in OpenTelemetry. As long as the maintainers can use and maintain the tooling they build in the language they're familiar with, it should be okay to submit PRs for tools in other languages. I may be missing your point so please clarify.

Oberon00:
>  Are you recommending all tooling for various languages supported in OpenTelemetry be developed in Python?

Python was never suggested in this discussion, only using the existing Jinja templating language. I certainly don't recommend all tools to be developed in Python. My issue was just that developing a dedicated tool for post-processing something that could just as easily be generated to the desired form by the previous step in the pipeline struck me as overly complicated. But I'm not blocking the PR based on this, I see your arguments of trying to minimize code in Jinja templates that the Go SIG can't maintain as well as Go code.

Oberon00:
> The name of the executable built from this import path is irrelevant and can be whatever the end user wants.

It's not irrelevant to me and it's mentioned in the readme as `semconvgen` which could mislead readers into thinking that this is the default semantic convention generator. Especially since the docker container is already called exactly that (`otel/semconvgen`).

PR #51, opened 2021-07-12
Semconvgen: Clean up requirements, build.

Oberon00:
Development and runtime requirements were mixed up. Also, we don't need the whole source tree in the docker container, let's just use a wheel.


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -500,7 +500,7 @@ def __str__(self):
 class TextWithLinks(str):
     parts: List[Union[str, MdLink]]
     raw_text: str
-    md_link = re.compile("\[([^\[\]]+)\]\(([^)]+)")
+    md_link = re.compile(r"\[([^\[\]]+)\]\(([^)]+)")
```
Oberon00, line 503: Unrelated warning fix.

carlosalberto (approved):

thisthat (approved):

arminru (approved):

Issue #47, opened 2021-06-15
Semantic  convention generator adds more than one blank line, failing mdlint


Oberon00:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1759#issuecomment-861645499

The PR in question has two enums in a single convention, and the first one has a note on one of the members. As a result, two consecutive newlines are printed between the footnote and the next enum table.

bryce-b:
I can take this ticket, but I don't think i'll be able to get a PR in until next week

Oberon00:
Just comment on the ticket if/when you start working on it.

bryce-b:
I'm working on it now, but I'm doing a half day today and I'm OOF tomorrow.

PR #52, opened 2021-07-13
Add markdownlint.

Oberon00:
Since we already had problems with the semantic convention generator generating markdown that failed the lint (#47) used in the spec, use the same linting rules here.


thisthat (approved):

.markdownlint.yaml
```diff
@@ -0,0 +1,14 @@
+# See https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
+# and https://github.com/DavidAnson/markdownlint/blob/main/README.md
+
+# Default state for all rules
```
arminru, line 4:
```suggestion
# This should be kept in sync with https://github.com/open-telemetry/opentelemetry-specification/blob/main/.markdownlint.yaml

# Default state for all rules
```
Maybe add a comment like this since the goal is to provide Markdown that is valid for the spec repo and hence use that in the test files.

arminru (approved):

carlosalberto (approved):

bogdandrutu (approved):

PR #53, opened 2021-07-14
Clean up docs, add syntax.md and semconv.schema.json

Oberon00:
As a follow-up I recommend removing the copy of `syntax.md` from https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/syntax.md and adapt the link in https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/README.md.

Logically, I think syntax.md needs to be here in the build-tools repository so we can update it in the same PR where we update something about the tool's input parsing (e.g. when adding new features).

### DIFF
--- a/semantic-conventions/README.md
+++ b/semantic-conventions/README.md
@@ -21,7 +21,8 @@ docker run --rm otel/semconvgen -h
 The expected YAML input file format is documented in [syntax.md](./syntax.md).
 
 There is also a JSON schema definition available for the YAML files, which can
-be used e.g. in VS code to get validation and auto-completion: [semconv.schema.json](semconv.schema.json). For example, with the `redhat.vscode-yaml` plugin, use the following snippet in your VS Code `settings.json` to apply it
+be used e.g. in VS code to get validation and auto-completion: [semconv.schema.json](./semconv.schema.json).
+For example, with the `redhat.vscode-yaml` plugin, use the following snippet in your VS Code `settings.json` to apply it
 to the test YAML files:
 
 ```json
@@ -49,7 +50,7 @@ Where `{yaml_folder}` is the absolute path to the directory containing the yaml 
 The tool will automatically replace the tables with the up to date definition of the semantic conventions.
 To do so, the tool looks for special tags in the markdown.
 
-```html
+```markdown
 <!-- semconv {semantic_convention_id} -->
 <!-- endsemconv -->
 ```
--- a/semantic-conventions/README.md
+++ b/semantic-conventions/README.md
@@ -26,7 +26,7 @@ docker run --rm otel/semconvgen --yaml-root {yaml_folder} markdown --markdown-ro
 
 Where `{yaml_folder}` is the absolute path to the directory containing the yaml files and
 `{markdown_folder}` the absolute path to the directory containing the markdown definitions
-(`specification` for [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/tree/master/)).
+(`specification` for [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/)).
 
 The tool will automatically replace the tables with the up to date definition of the semantic conventions.
 To do so, the tool looks for special tags in the markdown.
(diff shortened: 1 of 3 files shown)


Oberon00:
Gernot Reisinger wrote this as part of his job at Dynatrace, so I'll just go ahead and move him to Co-authored-by which is hopefully ignored by the CLA checker.

semantic-conventions/syntax.md
```diff
@@ -0,0 +1,235 @@
+# Semantic Convention YAML Language
```
Oberon00, line 1: As stated in the PR description, this file is copied from https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/syntax.md so that we can have any changes that affect the input format of the semantic convention generator in one PR in one repo, instead of split over spec and build-tools.

semantic-conventions/README.md
```diff
@@ -26,12 +44,12 @@ docker run --rm otel/semconvgen --yaml-root {yaml_folder} markdown --markdown-ro
...
 Where `{yaml_folder}` is the absolute path to the directory containing the yaml files and
 `{markdown_folder}` the absolute path to the directory containing the markdown definitions
-(`specification` for [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/tree/master/)).
+(`specification` for [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/tree/main/)).
 
 The tool will automatically replace the tables with the up to date definition of the semantic conventions.
 To do so, the tool looks for special tags in the markdown.
 
-```
+```html
```
arminru, line 52:
```suggestion
```markdown
```

semantic-conventions/README.md
```diff
@@ -16,6 +16,24 @@ For help try:
 docker run --rm otel/semconvgen -h
 ```
 
+## Model definition language (YAML input)
+
+The expected YAML input file format is documented in [syntax.md](./syntax.md).
+
+There is also a JSON schema definition available for the YAML files, which can
+be used e.g. in VS code to get validation and auto-completion: [semconv.schema.json](semconv.schema.json). For example, with the `redhat.vscode-yaml` plugin, use the following snippet in your VS Code `settings.json` to apply it
```
arminru, line 24:
```suggestion
be used e.g. in VS code to get validation and auto-completion: [semconv.schema.json](./semconv.schema.json).
For example, with the `redhat.vscode-yaml` plugin, use the following snippet in your VS Code `settings.json` to apply it
```

semantic-conventions/README.md
```diff
@@ -68,13 +86,27 @@ semantic conventions that have the tag `network`.
 
 The image supports [Jinja](https://jinja.palletsprojects.com/en/2.11.x/) templates to generate code from the models.
 
-For example, the following template is used by the [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/master/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/typedspan)
-to generate Java classes. [Template Link.](https://gist.github.com/thisthat/7e34742f4a7f1b5df57118f859a19c3b)
+For example, opentelemetry-java [generates typed constants for semantic conventions](https://github.com/open-telemetry/opentelemetry-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java)
+here using [this template](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2).
```
arminru, line 90:
```suggestion
using [this template](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2).
```

semantic-conventions/semconv.schema.json
```diff
@@ -0,0 +1,302 @@
+{
```
arminru, line 1: Have you verified that all YAML files that we have in the spec repo pass the validation?

semantic-conventions/syntax.md
arminru, line 1: Please add a note that when this file is changed one should also adapt `semconv.schema.json` accordingly.

semantic-conventions/semconv.schema.json
Oberon00: No, I don't know a tool that could do this validation. I'd need to open all files in VS code and look at the "problems" view I guess. The main purpose of the file is as an editing aid.

Oberon00: The semantic conventions are currently missing the `type:` property which should probably be fixed because at the moment the tool will then default to `span`, even for resources.

You can see warning messages if you look at the semantic convention checkers build output, e.g. here: https://github.com/open-telemetry/opentelemetry-specification/runs/3074608868

```
Using default SPAN type for semantic convention 'aws.ecs' @ line 5
Using default SPAN type for semantic convention 'aws.eks' @ line 5
Using default SPAN type for semantic convention 'aws.log' @ line 5
Using default SPAN type for semantic convention 'container' @ line 5
Using default SPAN type for semantic convention 'deployment' @ line 5
```

etc.

Oberon00: I checked all files (using VS code) and found no further problems in them. However, I then noticed a misspelling in the schema that caused something not to be checked, which I fixed. I also made the type optional to reflect reality (maybe we even need to remove `type` from the tool again if we want to share metric and span conventions). 583e29d

semantic-conventions/syntax.md
Oberon00: Done in f46daff, also added a TOC.

mariojonke (approved):

thisthat (approved):

Oberon00:
@open-telemetry/specs-approvers Please review if this seems to be the right direction to go (see PR description).

carlosalberto (approved):

PR #54, opened 2021-07-16
Fix semconv tests on Windows, refactor.

Oberon00:
Originally I only meant to fix a missing `encoding="utf-8"` in an `open()` call (that makes pytest fail on Windows where Python's default open()-encoding is not UTF-8), but seeing that it was duplicated in many places, I went ahead and refactored the unit test suite in question.


semantic-conventions/src/tests/semconv/templating/test_markdown.py
```diff
@@ -23,522 +25,139 @@
 
 class TestCorrectMarkdown(unittest.TestCase):
     def testRef(self):
-        semconv = SemanticConventionSet(debug=False)
-        semconv.parse(self.load_file("markdown/ref/general.yaml"))
-        semconv.parse(self.load_file("markdown/ref/rpc.yaml"))
-        semconv.finish()
-        self.assertEqual(len(semconv.models), 5)
```
Oberon00, line 30: We lose these assertions, but I think they are not very helpful anyway. This test suite is supposed to cover markdown rendering, not model parsing.

mariojonke (commented):
are [input_http.md](https://github.com/open-telemetry/build-tools/pull/54/files#diff-9d6c7c7df0742074446ca53dc6942755ffec42512b62ff19aea47063ec22a43e) and are [input_general.md](
https://github.com/open-telemetry/build-tools/pull/54/files#diff-7f8d5cf25eeb81821fd9d088341b46330d2b9fb4894c42f905e08dc0d4990e74) unused? there doesn't appear to be any test that required updating after moving these files to a different folder.

Oberon00:
These files seem to be referenced by the respective input.md (when the link is generated).

mariojonke (approved):

thisthat (approved):

arminru (approved):

PR #48, opened 2021-06-21
Fixes #47 multiple lines added in table generation

bryce-b:
also adds tests for said issue.

PR #55, opened 2021-07-20
Fix bug with newlines around enums.

Oberon00:
Turns out #48 merely hid the actual problem which was that the YAML parser adds a trailing newline if you use `>` strings. Probably that is correct, but not what we expect in this context.

In other places we were already using `.strip()`, so also use it here.

This caused some broken formatting internally at Dynatrace where we also use this tool.

This commit also affects the `code` command, not only markdown.


Oberon00:
I verified that the updated version does not change anything in the current spec, running `gen-semconv -f semantic_conventions\ markdown -md specification --md-check` in the spec yields `69 files left unchanged.`.

arminru (approved):

thisthat (approved):

PR #57, opened 2021-07-29
Add support for Event Semantic Conventions

Oberon00:
Event semantic conventions were designed with span events in mind but should be extendable to log events. See the new tests for how this can be useful to add a YAML definition for the exception semantic conventions in the spec.

This adds two new elements to the YAML definition:
* A new semantic convention type "event", with no additional fields
* A new additional field for "span" semantic conventions to reference associated event semantic conventions.
  The generator validates that the semantic convention exists and the new data is available to the code templates,
  but is not yet used in the markdown output.

It also updates the JSON schema to only allow span_kind and the new events field if the semantic convention type is not specified (defaults to span) or is "span". Unfortunately, we then lose the `"additionalProperties": false`, this seems to be a limitation of the JSON schema language, see https://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties#comment83207554_23001194

Full disclosure: This is the last feature for this tool we had developed internally at Dynatrace before (promise! 😄), so I'm starting off by submitting the status quo here in hopes our code bases back in sync (we already use the event semantic convention type internally).

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/version.py
+++ b/semantic-conventions/src/opentelemetry/semconv/version.py
@@ -12,4 +12,4 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-__version__ = "0.4.1"
+__version__ = "0.5.0"
--- a/semantic-conventions/src/tests/data/yaml/errors/events/missing_event.yaml
+++ b/semantic-conventions/src/tests/data/yaml/errors/events/missing_event.yaml
@@ -16,6 +16,6 @@ groups:
     prefix: random
     brief: example
     attributes:
-        - id: flag
-          type: boolean
-          brief: An attribute.
+      - id: flag
+        type: boolean
+        brief: An attribute.
--- a/semantic-conventions/src/tests/data/yaml/errors/events/missing_event.yaml
+++ b/semantic-conventions/src/tests/data/yaml/errors/events/missing_event.yaml
@@ -1,0 +1,21 @@
+groups:
+  - id: has.events
+    type: span
+    prefix: has.events
+    events:
+      - exception
+      - random.event
+    brief: example
+    attributes:
+      - id: flag
+        type: boolean
+        brief: An attribute.
+
+  - id: random.event
+    type: event
+    prefix: random
+    brief: example
+    attributes:
+        - id: flag
+          type: boolean
+          brief: An attribute.
--- a/semantic-conventions/src/tests/data/yaml/errors/events/no_event_type.yaml
+++ b/semantic-conventions/src/tests/data/yaml/errors/events/no_event_type.yaml
@@ -1,0 +1,20 @@
+groups:
+  - id: has.events
+    type: span
+    prefix: has.events
+    events:
+      - random.event
+    brief: example
+    attributes:
+      - id: flag
+        type: boolean
+        brief: An attribute.
+
+  - id: random.event
+    type: span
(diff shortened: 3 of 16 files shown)


semantic-conventions/src/opentelemetry/semconv/version.py
```diff
@@ -12,4 +12,4 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-__version__ = "0.4.1"
+__version__ = "0.5.0"
```
mariojonke, line 15: i guess the [changelog](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/CHANGELOG.md) should also be updated?

Oberon00: Oh, I did not realize we had one, yes it should.

Oberon00: 4720769

mariojonke (approved):

Oberon00:
@open-telemetry/specs-approvers this still needs reviews.

tigrannajaryan:
Maybe @open-telemetry/python-approvers can help review the Python portion and @open-telemetry/specs-logs-approvers can help review the event-related logic?

Oberon00:
The events here were designed for span events. But if something prevents them from being used as logs (other than missing features) that would be good to know.

semantic-conventions/src/tests/data/yaml/errors/events/missing_event.yaml
```diff
@@ -0,0 +1,21 @@
...
+        brief: An attribute.
+
+  - id: random.event
+    type: event
+    prefix: random
+    brief: example
+    attributes:
+        - id: flag
+          type: boolean
+          brief: An attribute.
```
aabmass, line 21: nit formatting

```suggestion
      - id: flag
        type: boolean
        brief: An attribute.
```

semantic-conventions/src/tests/data/yaml/errors/events/no_event_type.yaml
```diff
@@ -0,0 +1,20 @@
...
+        brief: An attribute.
+
+  - id: random.event
+    type: span
+    prefix: random
+    brief: example
+    attributes:
+        - id: flag
+          type: boolean
+          brief: An attribute.
```
aabmass, line 20: nit
```suggestion
      - id: flag
        type: boolean
        brief: An attribute.
```

semantic-conventions/src/tests/data/yaml/span_event.yaml
```diff
@@ -0,0 +1,21 @@
...
+        brief: An attribute.
+
+  - id: random.event
+    type: event
+    prefix: random
+    brief: example
+    attributes:
+        - id: flag
+          type: boolean
+          brief: An attribute.
```
aabmass, line 21:
```suggestion
      - id: flag
        type: boolean
        brief: An attribute.
```

aabmass (approved):
LGTM

semantic-conventions/src/tests/data/yaml/errors/events/no_event_type.yaml
Oberon00: Thanks! Maybe we should add yamllint to this repo too.

thisthat (approved):

PR #58, opened 2021-08-03
Check bool examples, don't crash on nonexistent examples.

Oberon00:
We don't run into this in the spec yet, but if we ever want to remove some integer examples, we will.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -252,7 +252,7 @@ class SemanticAttribute:
             raise ValidationError.from_yaml_pos(position, msg)
         if not isinstance(examples, CommentedSeq) and examples is not None:
             # TODO: If validation fails later, this will crash when trying to access position data
-            # since a list, contratry to a CommentedSeq, does not have position data
+            # since a list, contrary to a CommentedSeq, does not have position data
             examples = [examples]
         if is_simple_type and attr_type not in (
             "boolean",
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -251,7 +251,7 @@ class SemanticAttribute:
             msg = "Non array examples for {} are not allowed".format(attr_type)
             raise ValidationError.from_yaml_pos(position, msg)
         if not isinstance(examples, CommentedSeq) and examples is not None:
-            # TODO: If validation fails later, this will crash when trying to access positio data
+            # TODO: If validation fails later, this will crash when trying to access position data
             # since a list, contratry to a CommentedSeq, does not have position data
             examples = [examples]
         if is_simple_type and attr_type not in (
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -250,25 +250,25 @@ class SemanticAttribute:
             position = attribute.lc.data[list(attribute)[0]]
             msg = "Non array examples for {} are not allowed".format(attr_type)
             raise ValidationError.from_yaml_pos(position, msg)
-        if (
-            (is_simple_type or isinstance(attr_type, EnumAttributeType))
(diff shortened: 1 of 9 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -250,25 +250,25 @@ def parse_id(attribute):
...
             "double",
             "double[]",
-        ]:
-            if examples is None or (len(examples) == 0):
+        ):
+            if not examples:
                 position = attribute.lc.data[list(attribute)[0]]
                 msg = "Empty examples for {} are not allowed".format(attr_type)
                 raise ValidationError.from_yaml_pos(position, msg)
-        if is_simple_type and attr_type not in ["boolean", "boolean[]"]:
```
Oberon00, line 271: @thisthat Any idea why this was originally excluded? To support more boolean values like "yes" instead of "true"?

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -250,25 +250,25 @@ def parse_id(attribute):
             position = attribute.lc.data[list(attribute)[0]]
             msg = "Non array examples for {} are not allowed".format(attr_type)
             raise ValidationError.from_yaml_pos(position, msg)
-        if (
-            (is_simple_type or isinstance(attr_type, EnumAttributeType))
-            and not isinstance(examples, CommentedSeq)
-            and examples is not None
-        ):
+        if not isinstance(examples, CommentedSeq) and examples is not None:
+            # TODO: If validation fails later, this will crash when trying to access positio data
```
Oberon00, line 254: I'm under the impression that someone needs to read the YAML library documentation in more depth and refactor a lot of code (but not in this PR).

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
thisthat: I think we excluded boolean attributes from having an example because it is not very useful to have `true`/`false` there.

Oberon00: I know they are not required to have examples (though before this patch, non-array bools without examples would have crashed the generator). But in case they do have examples (which is allowed), the examples should still be checked for whether they have the right type.

mariojonke (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -250,25 +250,25 @@ def parse_id(attribute):
...
             msg = "Non array examples for {} are not allowed".format(attr_type)
             raise ValidationError.from_yaml_pos(position, msg)
-        if (
-            (is_simple_type or isinstance(attr_type, EnumAttributeType))
-            and not isinstance(examples, CommentedSeq)
-            and examples is not None
-        ):
+        if not isinstance(examples, CommentedSeq) and examples is not None:
+            # TODO: If validation fails later, this will crash when trying to access position data
+            # since a list, contratry to a CommentedSeq, does not have position data
```
arminru, line 255:
```suggestion
            # since a list, contrary to a CommentedSeq, does not have position data
```

semantic-conventions/src/tests/data/yaml/errors/examples/example_bool.yaml
```diff
@@ -9,4 +9,4 @@ groups:
         type: boolean
         required: always
         brief: 'test'
-        examples: [true, false, yes]
+        examples: [true, false, true]
```
arminru, line 12: Those examples don't really make sense. You could instead check that `yes` and `1` are rejected as expected.

Oberon00: I added a new test. It is possible that this test was supposed to work differently, since it is in the errors/ directory, but the accompanying test does not assert on anything.

Oberon00: You are right, the test was wrong, and the YAML was intentionally invalid. Fixed in 1272f57, 
e205225

arminru (approved):

mariojonke (approved):

Issue #63, opened 2021-09-08
Semantic convention generator: Clean up tests using separate input files


Oberon00:
Many tests that use separate input files are not written very cleanly. The most important points:

* The input YAML files contain lots of redundant stuff (copied from real semantic conventions). They should be minimized to only contain what is needed for the test.
* The Python code of the tests could use some cleanup too similar to what was already done for some markdown tests, see https://github.com/open-telemetry/build-tools/pull/54/files#diff-6efabfcf3a7471d21e99e83685a2d0151cbb1404c52eed86f26a06d81ef17c44 where the test python code was cut down from 544 to 153 lines.
* Instead of or in addition to the above point, try testing in such a way that we can use Python (triple-quoted) strings as input instead of needing separate files. This is not entirely trivial since file names do play a role (for generating links), but should nonetheless be doable. This would make the tests much, much more readable.

Further ideas, not as important:

* Asserting on the line in which an exception occurs could be cleaner if we could somehow specify the expected exception in the input (e.g. with a YAML/HTML comment).

Oberon00:
See also https://github.com/open-telemetry/build-tools/pull/79#discussion_r851360604:

> These tests are not designed as integration tests, they are targeted feature tests and it makes no sense to drag in a full spec yaml. These tests should instead use a minimal yaml that just uses group foo with attribute bar instead of actual data. See https://github.com/open-telemetry/build-tools/issues/63.
>
> Having tests that automatically pull in spec code would additionally make sense, but then we don't have e.g. a testMultipleEnum or testRef but instead a testGeneratorIsSpecCompatible (and you'd need to think about how to design the build/change process so that you can also change the output in "incompatible" ways, e.g. temporarily reference a spec PR branch that updates the spec so the semconv table check is in sync again.

lmolkova:
closing this issue - we're sunsetting otel/semconvgen in favor of [weaver](https://github.com/open-telemetry/weaver) https://github.com/open-telemetry/build-tools/issues/322 and won't make any improvements here.

PR #64, opened 2021-09-08
Validate that enum member IDs are valid

### LINKED ISSUE #59: to_const_name generates invalid variable identifiers
Recently in semantic conventions, two new types of id were introduced:

- Alibaba Cloud (https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/cloud.yaml#L11)
- 1xRTT (https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/general.yaml#L94)

The const names generated for those ids are `ALIBABA CLOUD` and `1XRTT` respectively. Those generated values are not valid variable identifiers in most languages.

In opentelemetry-js, we use `to_const_name` to generate keys for those ids: https://github.com/open-telemetry/opentelemetry-js/blob/main/scripts/semconv/templates/SemanticAttributes.ts.j2#L57. In this case, the keys should be valid variable identifiers.

Should we change the semantic conventions or update the generator to generate valid identifiers?

Oberon00:
Validate that enum member IDs are valid, improve error messages (fixes #59).

We already have ID validation, but enum member IDs were missing.

I ran this against the current semantic conventions from spec, and there are no issues.

Sorry for the tons of changed test files, this is currently unavoidable, but I opened issue #63 to improve the situation in the future. Try filtering out yaml and md files using the GitHub UI.


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -432,32 +434,39 @@ def parse(attribute_type):
...
         for member in attribute_type["members"]:
             validate_values(member, allowed_keys, mandatory_keys)
             if not EnumAttributeType.is_valid_enum_value(member["value"]):
-                raise ValidationError(
-                    0, 0, "Invalid value used in enum: <{}>".format(member["value"])
+                raise ValidationError.from_yaml_pos(
+                    member.lc.data["value"][:2],
+                    "Invalid value used in enum: <{}>".format(member["value"]),
                 )
+            validate_id(member["id"], member.lc.data["id"])
```
Oberon00, line 454: This is the most relevant line in this PR.

mariojonke (approved):

semantic-conventions/src/opentelemetry/semconv/model/utils.py
```diff
@@ -17,15 +17,17 @@
 
 from opentelemetry.semconv.model.exceptions import ValidationError
 
-
 ID_RE = re.compile("([a-z](\\.?[a-z0-9_-]+)+)")
+"""Identifiers must start with a lowercase ASCII letter and
+contain only lowercase, digits 0-9, underscore and dots.
```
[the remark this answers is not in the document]
Oberon00: Oh, you are right. That's curious. @thisthat Do you remember the reason for that?
Anyway, `to_const_name`handles that by replacing with underscores so it shouldn't be a big deal.

```python
def to_const_name(name: str) -> str:
    return name.upper().replace(".", "_").replace("-", "_")
```

Oberon00, line 22:
```suggestion
contain only lowercase, digits 0-9, underscore, dash (not recommended) and dots.
```

semantic-conventions/src/opentelemetry/semconv/version.py
```diff
@@ -12,4 +12,4 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-__version__ = "0.5.0"
+__version__ = "0.6.0"
```
arminru, line 15: Also update `CHANGELOG.md` please.

arminru (approved):

thisthat (approved):

semantic-conventions/src/opentelemetry/semconv/model/utils.py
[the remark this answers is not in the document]
Oberon00: It may make sense if we use `-` for HTTP headers and want to call out specific headers... Although the current PR for that was changed to normalize to `_`, see https://github.com/open-telemetry/opentelemetry-specification/pull/1898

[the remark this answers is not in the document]
Oberon00: Anyway, not really related to this PR, so I'm resolving this conversation.

semantic-conventions/src/opentelemetry/semconv/version.py
Oberon00: Done in 48a41b.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
Oberon00: "Resolving" this because it apparently now blocks merging.

PR #62, opened 2021-09-07
Run & fix linters flake8, pylint, mypy

Oberon00:
This fixes errors from & runs the following linters, that are all also used in opentelemetry-python:

- flake8 (a fast & simple linter)
- pylint (a slow linter that does some code analysis)
- mypy (a checker for the type annotations we already had in place)

PR #65, opened 2021-09-15
Add test against setting attributes for units.

Oberon00:
This is something I missed with removing the `HasAttributes` base class in #62, but fortunately it works as it should as this test demonstrates (because `UnitSemanticConvention` explicitly sets `allowed_keys` to something not containing `attributes`).


semantic-conventions/src/tests/semconv/model/test_semantic_convention_units.py
```diff
@@ -1,9 +1,12 @@
 import os
+from opentelemetry.semconv.model.exceptions import ValidationError
 from opentelemetry.semconv.model.semantic_convention import (
     parse_semantic_convention_groups,
     UnitSemanticConvention,
 )
 
+import pytest
```
mariojonke, line 8: will probably conflict with isort rules once https://github.com/open-telemetry/build-tools/pull/66/files gets merged

mariojonke (approved):

semantic-conventions/src/tests/semconv/model/test_semantic_convention_units.py
Oberon00: Rebased.

arminru (approved):

PR #67, opened 2021-09-15
Add event name field, fix leading dot in FQN of attribute w/o prefix.

Oberon00:
All event semantic conventions now have a `name: str` attribute. This defaults to the `prefix`, but if no prefix is given or a different name is desired, can be manually set with the `name` property in the YAML.

The markdown generator adds a single line before the table for event semantic conventions, see semantic-conventions/src/tests/data/markdown/event/expected.md

>  The event name MUST be `myname`.

Events with dynamic names are not supported, as there is currently no use case in the spec.

CC @bogdandrutu

This needs a spec follow-up, I created https://github.com/open-telemetry/opentelemetry-specification/pull/1922


semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -117,7 +117,10 @@ def parse(
             if attr_id is not None:
                 validate_id(attr_id, position_data["id"])
                 attr_type, brief, examples = SemanticAttribute.parse_id(attribute)
-                fqn = "{}.{}".format(prefix, attr_id)
+                if prefix:
```
Oberon00, line 120: I did not add this bugfix to the changelog because IMHO the prefix being optional is a misfeature, and we should require the prefix and/or default it to the semantic convention ID.

mariojonke (approved):

bogdandrutu (approved):
The semantics LGTM, not at all a Python expert :)

arminru (approved):

PR #68, opened 2021-09-15
Support sampling_relevant attributes

lmolkova:
Related to https://github.com/open-telemetry/opentelemetry-specification/pull/1916

### DIFF
--- a/semantic-conventions/syntax.md
+++ b/semantic-conventions/syntax.md
@@ -7,17 +7,20 @@ Then, the semantic of each field is described.
 
 <!-- toc -->
 
-- [JSON Schema](#json-schema)
-- [Syntax](#syntax)
-- [Semantics](#semantics)
-  * [Groups](#groups)
-  * [Semantic Convention](#semantic-convention)
-  * [Attributes](#attributes)
-  * [Ref](#ref)
-  * [Type](#type)
-  * [Constraints](#constraints)
-    + [Any Of](#any-of)
-    + [Include](#include)
+- [Semantic Convention YAML Language](#semantic-convention-yaml-language)
+  - [JSON Schema](#json-schema)
+  - [Syntax](#syntax)
+  - [Semantics](#semantics)
+    - [Groups](#groups)
+    - [Semantic Convention](#semantic-convention)
+      - [Span semantic convention](#span-semantic-convention)
+    - [Attributes](#attributes)
+      - [Examples (for examples)](#examples-for-examples)
+      - [Ref](#ref)
+      - [Type](#type)
+    - [Constraints](#constraints)
+      - [Any Of](#any-of)
+      - [Include](#include)
 
 <!-- tocstop -->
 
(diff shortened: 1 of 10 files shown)


semantic-conventions/syntax.md
```diff
@@ -7,17 +7,20 @@ Then, the semantic of each field is described.
 
 <!-- toc -->
 
-- [JSON Schema](#json-schema)
```
Oberon00, line 10: I think there is no need to regenerate the TOC.

Oberon00:
It looks like you need to run the `black` automformatter again. Try `pip install -r dev-requirements.txt` and then `black .`

Oberon00:
Please add a test to https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/tests/semconv/templating/test_markdown.py (plus accompanying input.md, expected.md and minimal yaml file with any name). Then this LGTM 👍

semantic-conventions/syntax.md
lmolkova: vscode auto-regenerates it for me. Anyway, returned original one back

lmolkova:
> Please add a test to https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/tests/semconv/templating/test_markdown.py (plus accompanying input.md, expected.md and minimal yaml file with any name). Then this LGTM 👍

I don't know when I'll get a chance to work on this. Simple spec update turned out to be way too complicated.

lmolkova:
> > Please add a test to https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/tests/semconv/templating/test_markdown.py (plus accompanying input.md, expected.md and minimal yaml file with any name). Then this LGTM 👍
> 

Done!

semantic-conventions/syntax.md
```diff
@@ -174,7 +174,6 @@ An attribute is defined by:
 - `sampling_relevant`, optional EXPERIMENTAL boolean,
```
mariojonke, line 174: is it still considered experimental?

mariojonke (commented):
[changelog](https://github.com/open-telemetry/build-tools/blob/v0.6.0/semantic-conventions/CHANGELOG.md) and [version.py](https://github.com/open-telemetry/build-tools/blob/v0.6.0/semantic-conventions/src/opentelemetry/semconv/version.py) should probably be updated.

Oberon00:
@mariojonke I can handle that in #67 or a follow-up.

Oberon00 (approved):
Thank you!

mariojonke (approved):

semantic-conventions/syntax.md
lmolkova: I don't know, but they can stay experimental as far as I am concerned - all semantic conventions are experimental.

lmolkova:
I updated the changelog and version. If it still looks good, can someone please merge it? Thanks!

arminru:
@open-telemetry/specs-approvers This will affect the semantic conventions we have in the spec, please take a look.

lmolkova:
> @open-telemetry/specs-approvers This will affect the semantic conventions we have in the spec, please take a look.

Current semantic conventions don't use `sampling_relevant` field, i.e. it does not change any of them.

Here's the change where I start using `sampling_relevant`, which is approved and waiting for this PR to fix the build.

lmolkova:
@Oberon00 @arminru @mariojonke can someone please merge it?

Oberon00:
Only the TC can merge PRS in this repo, same as the spec (so @arminru _can_ merge). IMHO we can merge this, because as you said the current spec does not use sampling_relevant.

arminru:
Fair enough as the actual usage of it will still be reviewed in the spec.
Released as https://github.com/open-telemetry/build-tools/releases/tag/v0.7.0 so `otel/semconvgen:0.7.0` is ready for you to reference in your spec PR @lmolkova.

lmolkova:
thank you all!

PR #73, opened 2021-10-08
Build multiarch protobuf Docker images

kvrhdn:
This adapts `protobuf/Dockerfile` so it can be built for amd64 and arm64 targets by using `TARGETARCH`.
I had to update `PROTOC_GEN_LINT_VERSION` as only [v0.2.4](https://github.com/ckaznocha/protoc-gen-lint/releases) has downloads for all archs.

Next I updated the GitHub Actions workflow, I'm using [docker/build-push-action](https://github.com/docker/build-push-action) to build multiarch images. Under the hood this uses docker buildx.

You can run buildx locally using:
```
docker buildx build --load --platform linux/arm64 -t otel/protobuf:latest protobuf
```

Fixes https://github.com/open-telemetry/build-tools/issues/42

PR #75, opened 2021-10-20
make it possible to build the proto compiler using arm64

codeboten:
The default is still to use `amd64` for all builds, but this makes it possible for someone to manually build the proto generator locally. Next step is to add a step to the build to publish a docker container that supports arm64


jmacd (approved):

codeboten:
Closing in favour of  #73

PR #83, opened 2022-02-15
Add schema file tool

tigrannajaryan:
This can be used in specification repo to validate the schema files.

### DIFF
--- a/.github/workflows/schema_tools.yml
+++ b/.github/workflows/schema_tools.yml
@@ -1,0 +1,41 @@
+name: Schema File Tools Docker Image
+on:
+  push:
+    tags: [ '**' ]
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    paths:
+      - .github/workflows/schema_tools.yml
+      - schemas/*
+
+jobs:
+  build:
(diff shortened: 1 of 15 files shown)


tigrannajaryan:
@open-telemetry/specs-approvers please review.

SergeyKanzhelev (approved):
lgtm for implemented checks. Curious do you have any other checks in mind?

Also, adding unit tests would be a nice addition demonstrating what the tool does

tigrannajaryan:
> lgtm for implemented checks. Curious do you have any other checks in mind?

Yes, it is possible to add more checks (e.g. that the historical versions agree with each other). This is just the first step, we can build on top of it.

> 
> Also, adding unit tests would be a nice addition demonstrating what the tool does

Good point. 

This primarily uses the schema parser from Go SDK which already has some unit tests: https://github.com/open-telemetry/opentelemetry-go/blob/main/schema/v1.0/parser_test.go

Let me see if I can add any useful tests here as well.

tigrannajaryan:
Added a few tests.

tigrannajaryan:
Merging to start using in the spec. More improvements can be in future PRs.

PR #85, opened 2022-02-23
Improve wording for custom values

trask:
based on feedback https://github.com/open-telemetry/opentelemetry-specification/pull/2377/files#r813153678

### DIFF
--- a/semantic-conventions/src/tests/data/markdown/deprecated/expected.md
+++ b/semantic-conventions/src/tests/data/markdown/deprecated/expected.md
@@ -16,7 +16,7 @@
 
 **[1]:** If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
-The following values are defined for `http.flavor`. If one of them applies, then it MUST be used, otherwise a custom value MAY be used.
+`http.flavor` has the following list of well-known values. If one of them applies, then it MUST be used, otherwise a custom value MAY be used.
 
 | Value  | Description |
 |---|---|
--- a/semantic-conventions/src/tests/data/markdown/deprecated/expected.md
+++ b/semantic-conventions/src/tests/data/markdown/deprecated/expected.md
@@ -16,7 +16,7 @@
 
 **[1]:** If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
-`http.flavor` If one of the listed values applies, then the respective value MUST be used, otherwise a custom value MAY be used:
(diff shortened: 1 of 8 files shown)


reyang (approved):

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
```diff
@@ -16,7 +16,7 @@
 
 **[1]:** If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
-`http.flavor` MUST be one of the following or, if none of the listed values apply, a custom value:
+`http.flavor` If one of the listed values applies, then the respective value MUST be used, otherwise a custom value MAY be used:
```
arminru, line 19: This reads quite strange. Before the change it was a "natural" sentence but now it's just appended to the name of the attribute.

How about

> The following values are defined for `http.flavor`. If one of them applies, then ...

or at least adding a colon

> `http.flavor`: If one of the values listed below applies, then the respective value MUST be used, otherwise a custom value MAY be used:

trask: makes sense, ptal

semantic-conventions/src/tests/data/markdown/deprecated/expected.md
```diff
@@ -16,7 +16,7 @@
 
 **[1]:** If `net.transport` is not specified, it can be assumed to be `IP.TCP` except if `http.flavor` is `QUIC`, in which case `IP.UDP` is assumed.
 
-`http.flavor` MUST be one of the following or, if none of the listed values apply, a custom value:
+The following values are defined for `http.flavor`. If one of them applies, then it MUST be used, otherwise a custom value MAY be used.
```
Oberon00, line 19: I think I would still like to have the name of the attribute being described stand out more.

One suggestion is to add a section to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md or a similar place to explain the concept of "open enums" (name up for debate). Then we could write "`http.flavor` is an open enum (<-- link to definition here) with the following non-exhaustive list of possible values:"


Or we could simply do:

```suggestion
`http.flavor` SHOULD be set to one of the following values:
```

And for allow_custom_values=false, we use MUST instead of should.

If you keep it like it is now with the attribute name not at the start, I suggest to at least make the attribute name bold by adding `**` around it.

trask: ptal, it now reads:

`http.flavor` has the following list of well-known values. If one of them applies, then it MUST be used, otherwise a custom value MAY be used.

Oberon00 (approved):

arminru:
Please adapt the test data to the latest changes and then this is good to merge from my end. Thanks @trask!

bogdandrutu (approved):

PR #87, opened 2022-02-23
Fix semantic-conventions build

trask:
It is currently failing (see build failure in #85) due to https://github.com/pallets/markupsafe/issues/282


tigrannajaryan (approved):

tigrannajaryan:
Thanks for the fix!

PR #102, opened 2022-06-28
Apply requirement level and msg of referenced attribute by default

lmolkova:
If the requirement level is not overridden when referencing an attribute, we should reuse the original one

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -461,7 +461,8 @@ class SemanticConventionSet:
                     attr.brief = ref_attr.brief
                 if not attr.requirement_level:
                     attr.requirement_level = ref_attr.requirement_level
-                    attr.requirement_level_msg = ref_attr.requirement_level_msg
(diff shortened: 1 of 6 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -459,6 +459,10 @@ def resolve_ref(self, semconv):
                 attr.attr_type = ref_attr.attr_type
                 if not attr.brief:
                     attr.brief = ref_attr.brief
+                if not attr.requirement_level:
+                    attr.requirement_level = ref_attr.requirement_level
+                if not attr.requirement_level_msg:
```
Oberon00, line 464: I think the message should only be inherited if the requirement_level is not overriden.

Oberon00:
Did you check if anything unintentionally changes in the generated semantic convention markdown, if using this branch?

lmolkova:
> Did you check if anything unintentionally changes in the generated semantic convention markdown, if using this branch?

yes, I did - there are no changes - we never specified the requirement level for common attributes

Oberon00 (approved):

arminru (approved):

PR #107, opened 2022-07-22
Remove noise

CleanCut:
This code emits meaningless noise.


Oberon00:
This is not meaningless noise. It warns you that you should add the `type:`  field to your semantic convention group, which is documented in https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#semantic-convention

Oberon00 (changes requested):
This is not meaningless noise. It warns you that you should add the type property to your semantic convention group in the YAML file, which is documented in https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#semantic-convention

arminru:
Superseded by https://github.com/open-telemetry/build-tools/pull/109 and https://github.com/open-telemetry/opentelemetry-specification/pull/2693.

PR #114, opened 2022-07-29
Add scope semantic conventions as a type

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-specification/issues/2682

Addresses this comment: https://github.com/open-telemetry/opentelemetry-specification/pull/2702#discussion_r933096178

This adds "scope" as a semantic convention type.

cc @joaopgrassi

### DIFF
--- a/semantic-conventions/CHANGELOG.md
+++ b/semantic-conventions/CHANGELOG.md
@@ -4,7 +4,8 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
-- ...
+- Add "scope" as a span type
+  ([#114](https://github.com/open-telemetry/build-tools/pull/114)).
 
 ## v0.13.0
 
--- a/semantic-conventions/CHANGELOG.md
+++ b/semantic-conventions/CHANGELOG.md
@@ -4,7 +4,7 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
-- Add "scope" as a span type
+- Add a semantic convention type for Instrumentation Scope ("scope")
   ([#114](https://github.com/open-telemetry/build-tools/pull/114)).
 
 ## v0.13.0
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -194,6 +194,10 @@ class ResourceSemanticConvention(BaseSemanticConvention):
     GROUP_TYPE_NAME = "resource"
 
 
+class ScopeSemanticConvention(BaseSemanticConvention):
+    GROUP_TYPE_NAME = "scope"
+
+
 class SpanSemanticConvention(BaseSemanticConvention):
     GROUP_TYPE_NAME = "span"
 
@@ -541,5 +545,6 @@ CONVENTION_CLS_BY_GROUP_TYPE = {
         EventSemanticConvention,
         MetricSemanticConvention,
         UnitSemanticConvention,
+        ScopeSemanticConvention,
     )
 }
--- a/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
+++ b/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
@@ -1,6 +1,6 @@
 groups:
   - id: scope
-    prefix:
+    prefix: ""
     type: scope
     brief: >
       Instrumentation Scope attributes
--- a/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
+++ b/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
@@ -1,6 +1,7 @@
 groups:
   - id: scope
-    prefix: scope
+    prefix:
+    type: scope
     brief: >
       Instrumentation Scope attributes
     attributes:
--- a/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
+++ b/semantic-conventions/src/tests/data/markdown/scope/scope.yaml
@@ -1,0 +1,12 @@
+groups:
+  - id: scope
+    prefix: scope
+    brief: >
+      Instrumentation Scope attributes
+    attributes:
+      - id: short_name
+        type: string
+        requirement_level: recommended
+        brief: >
+          The single-word name for the instrumentation scope.
+        examples: ['mylibrary']
--- a/semantic-conventions/src/tests/data/yaml/scope.yaml
+++ b/semantic-conventions/src/tests/data/yaml/scope.yaml
@@ -1,6 +1,6 @@
 groups:
   - id: scope-id
-    prefix: scopeprefix
+    prefix:
     type: scope
     brief: >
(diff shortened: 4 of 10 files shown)


semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -194,6 +194,10 @@ class ResourceSemanticConvention(BaseSemanticConvention):
     GROUP_TYPE_NAME = "resource"
 
 
+class ScopeSemanticConvention(BaseSemanticConvention):
+    GROUP_TYPE_NAME = "scope"
```
arminru, line 198: Please also document the new type here: https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#syntax
And add an example for it in a unit test.
Thanks!

joaopgrassi: You can add a test using the same `yaml` and `.md` from your PR. One for the parsing as in https://github.com/open-telemetry/build-tools/blob/f8714186e10f6bb9fe2f92ada2f3d23760ebed46/semantic-conventions/src/tests/semconv/model/test_correct_parse.py#L55 and maybe another for the markdown generation as in https://github.com/open-telemetry/build-tools/blob/f8714186e10f6bb9fe2f92ada2f3d23760ebed46/semantic-conventions/src/tests/semconv/templating/test_markdown.py#L117

semantic-conventions/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
-- ...
+- Add "scope" as a span type
```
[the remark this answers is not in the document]
joaopgrassi: Oh I see @arminru  also suggested a change. Either is fine by me :)

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
dashpole: added the syntax, and the unit test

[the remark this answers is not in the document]
dashpole: done

semantic-conventions/src/tests/semconv/templating/test_markdown.py
```diff
@@ -126,6 +126,9 @@ def test_event_renamed(self):
     def testSamplingRelevant(self):
         self.check("markdown/sampling_relevant/")
 
+    def testScope(self):
```
joaopgrassi, line 129:
```suggestion
    def test_scope(self):
```
Maybe?

joaopgrassi (approved):
Looks good to me! Adding @Oberon00 as he has more experience with the tool in general.

semantic-conventions/src/tests/data/yaml/scope.yaml
```diff
@@ -0,0 +1,12 @@
+groups:
+  - id: scope
+    prefix: scope
+    brief: >
+      Instrumentation Scope attributes
+    attributes:
```
arminru, line 6: If you don't set any `type: scope` here it will just fall back to the default span (but still work). This way the test doesn't cover your newly introduced type at all, however.

semantic-conventions/src/tests/semconv/model/test_correct_parse.py
```diff
@@ -669,6 +669,23 @@ def semantic_convention_check(self, s, expected):
...
         self.assertEqual(expected["attributes"], [a.fqn for a in s.attributes])
 
+    def test_scope_attribute(self):
+        semconv = SemanticConventionSet(debug=False)
+        semconv.parse(self.load_file("yaml/scope.yaml"))
+        self.assertEqual(len(semconv.models), 1)
+
+        expected = {
+            "id": "scope",
+            "prefix": "scope",
```
arminru, line 679: I think you should also assert for the `type` to be `scope` here.
Ideally also use some dummy names for id and prefix - if it's always `scope` in all occasions the test would not be able to discover any mix-up of the different fields.

dashpole: done

semantic-conventions/src/tests/data/markdown/scope/scope.yaml
```diff
@@ -0,0 +1,12 @@
+groups:
+  - id: scope
+    prefix: scope
+    brief: >
```
arminru, line 4:
```suggestion
    type: scope
    brief: >
```
Same as https://github.com/open-telemetry/build-tools/pull/114#discussion_r935327622.

arminru (approved):
Thank you, @dashpole!

joaopgrassi (approved):

joaopgrassi:
@Oberon00, @dashpole had to make a small change to allow the `prefix` to be empty, in order to generate the attribute without it. Other than that, I think this looks good!

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -123,7 +123,10 @@ def __init__(self, group):
 
         self.semconv_id = self.id
         self.note = group.get("note", "").strip()
-        self.prefix = group.get("prefix", "").strip()
+        if group.get("prefix", "") is None:
```
Oberon00, line 126: I don't think this can ever be None? Except if you actually write `null` in the YAML. I.e., I don't understand why this was required. What was the error?

dashpole: Workflow run: https://github.com/open-telemetry/build-tools/runs/7694090819?check_suite_focus=true

The error was:
```
        self.semconv_id = self.id
        self.note = group.get("note", "").strip()
>       self.prefix = group.get("prefix", "").strip()
E       AttributeError: 'NoneType' object has no attribute 'strip'
```

Oberon00: Maybe in YAML an empty attribute is equivalent to setting it to `null`? Can you try `prefix: ""` in the YAML?

dashpole: Done.  That works.

Oberon00: It's always surprising how surprising a "simple" YAML file can behave...

Oberon00 (approved):

arminru (approved):
Thank you, @dashpole!

dashpole:
What needs to be done to update this on my PR (so it passes tests)? https://github.com/open-telemetry/opentelemetry-specification/pull/2702

joaopgrassi:
@dashpole we need a new release/image of the tools and then update the makefile in the spec repo https://github.com/open-telemetry/opentelemetry-specification/blob/main/Makefile#L11. That should do it.

PR #116, opened 2022-08-08
Build multiarch docker images on tag

### LINKED ISSUE #115: Update otel/build-protobuf for multi-architecture support
Unable to build the opentelemetry-go-proto release for OTLP v0.19.0 on an arm chip because the build tools are for amd64.

The release steps fail like so:
```
% make sync VERSION=v0.19.0
upgrading opentelemetry-proto submodule to v0.19.0
HEAD is now at 6459e1a Prepare for v0.19.0 release (#420)
rm -rf gen otlp
rm -rf ./gen/go
mkdir -p ./gen/go
docker run --rm -u 502 -v/Users/josh.macdonald/src/opentelemetry/proto-go:/Users/josh.macdonald/src/opentelemetry/proto-go -w/Users/josh.macdonald/src/opentelemetry/proto-go otel/build-protobuf:0.11.0 --proto_path="gen/proto"  --go_out=./gen/go gen/proto/opentelemetry/proto/common/v1/common.proto
Unable to find image 'otel/build-protobuf:0.11.0' locally
0.11.0: Pulling from otel/build-protobuf
Digest: sha256:a1f16b31cb70dca3e486afad20c268cf9d3ee776b0cd6656120c8e013c27a52c
Status: Downloaded newer image for otel/build-protobuf:0.11.0
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
--go_out: protoc-gen-go: Plugin killed by signal 11.
make: *** [gen-otlp-protobuf] Error 1
```

MikeGoldsmith:
Updates the github workflow to use standard docker actions to build and publish multi-arch images when a git tag is pushed.

The workflow creates two tags, `latest` and one that is the git tag with the leading v stripped off. For example, the tag `v1.2.3` would create an image with tags `latest` and `1.2.3`.

- Closes #115
- Closes https://github.com/open-telemetry/opentelemetry-proto-go/issues/77

### DIFF
--- a/.github/workflows/protobuf-dockerimage.yml
+++ b/.github/workflows/protobuf-dockerimage.yml
@@ -14,8 +14,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-
     # https://github.com/docker/setup-qemu-action
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
@@ -50,7 +48,6 @@ jobs:
     - name: Build and push
       uses: docker/build-push-action@v2
       with:
-        context: .
         file: ./protobuf/Dockerfile
         platforms: linux/amd64,linux/arm64
         push: ${{ github.event_name != 'pull_request' }}
--- a/.github/workflows/protobuf-dockerimage.yml
+++ b/.github/workflows/protobuf-dockerimage.yml
@@ -52,7 +52,7 @@ jobs:
       uses: docker/build-push-action@v2
       with:
         context: "{{defaultContext}}:protobuf"
-        platforms: linux/amd64
+        platforms: linux/arm64
         push: ${{ github.event_name != 'pull_request' }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/protobuf-dockerimage.yml
+++ b/.github/workflows/protobuf-dockerimage.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Build and push
       uses: docker/build-push-action@v2
       with:
-        file: ./protobuf/Dockerfile
+        file: protobuf/Dockerfile
         platforms: linux/amd64,linux/arm64
         push: ${{ github.event_name != 'pull_request' }}
         tags: ${{ steps.meta.outputs.tags }}
--- a/.github/workflows/protobuf-dockerimage.yml
+++ b/.github/workflows/protobuf-dockerimage.yml
@@ -25,14 +25,6 @@ jobs:
       id: buildx
       uses: docker/setup-buildx-action@v1
 
-    # https://github.com/docker/login-action
-    - name: Login to DockerHub
-      if: github.event_name != 'pull_request'
-      uses: docker/login-action@v1
-      with:
-        username: ${{ secrets.DOCKER_USERNAME }}
-        password: ${{ secrets.DOCKER_PASSWORD }}
-
     # https://github.com/docker/metadata-action
     - name: Docker meta
       id: meta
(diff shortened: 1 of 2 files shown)


jmacd (approved):

.github/workflows/protobuf-dockerimage.yml
```diff
@@ -15,20 +15,45 @@ jobs:
...
+    - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            otel/build-protobuf
+          # generate Docker tags based on the following events/attributes
+          tags: |
+            type=schedule
```
arminru, line 46: Does this apply here as the workflow only runs on push and PR?

MikeGoldsmith: Yes, this isn't necessary and can be removed 👍🏻

MikeGoldsmith:
Okay, fixed the docker pathing issue. Now it's failing on installing protoc_gen_lint [here](https://github.com/open-telemetry/build-tools/blob/main/protobuf/Dockerfile#L131) - this runs fine locally for me.

MikeGoldsmith:
Okay, so I've confirmed that the TARGETARCH build arg isn't getting set by the buildx command as expected - see [here](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/).

I'm working on a way to detect the target arch and set the arg another way.

This works when run and setting the build arg manually like this `docker build protobuf/. -t build-tools --build-arg TARGETARCH=arm64`.

MikeGoldsmith:
Well, it looks like the problem was that the generic top level ARG default was overriding the ones passed in. After removing, it seems to have worked. However, the build is now super at more than 2 hours. This is likely because we're using QEMU and compiling our own grpc binaries.

Maybe we could source prebuilt binaries from somewhere else?

jmacd:
I'm not sure what we should do. Practically speaking, it's easy enough for me to find an amd64 machine to build with. 🤣

jmacd:
@MikeGoldsmith do you mean that the build time is 2 hours _to build the docker images_? That doesn't sound like a problem, and it would be great to get this merged.

jmacd (changes requested):
I've tried to use the docker images built using this branch and it seems we're not quite ready to build arm64 binaries using these tools.  The `grpc_cpp_plugin` segfaults:
```
# ./grpc_cpp_plugin
Segmentation fault
```
That was the only one of the following list that segfaults:
```
# ls -l *plugin*
-rwxr-xr-x    1 root     root         31352 Sep 19 17:35 grpc_cpp_plugin
-rwxr-xr-x    1 root     root         27912 Sep 19 17:35 grpc_csharp_plugin
-rwxr-xr-x    1 root     root         27384 Sep 19 17:35 grpc_node_plugin
-rwxr-xr-x    1 root     root         22736 Sep 19 17:35 grpc_objective_c_plugin
-rwxr-xr-x    1 root     root         32960 Sep 19 17:35 grpc_php_plugin
-rwxr-xr-x    1 root     root         16512 Sep 19 17:35 grpc_python_plugin
-rwxr-xr-x    1 root     root         22520 Sep 19 17:35 grpc_ruby_plugin
```

jmacd:
I was mistaken, the `grpc_objective_c_plugin` also segfaults.

jmacd:
How to repeat the results quoted above:

In this branch, I ran:

```
docker build protobuf/. -t build-tools --build-arg TARGETARCH=arm64 
docker tag build-tools:latest build-tools:multiarch
```

Then, in an otel-proto repository:

```
make OTEL_DOCKER_PROTOBUF=build-tools:multiarch
```

lucacome:
I was about to open a similar PR (didn't notice this one)...is anybody still working on this one?

Issue #120, opened 2022-09-26
Make diffs in table regeneration issues more visible


hughesjj:
I (and apparently another engineer) had trouble seeing why exactly the build process was failing, the error message only being 

```
File {blah} contains a table that would be reformatted.
```

I've since found the tool is very particular about whitespace and even the number of dashes in a table header.

Oberon00:
For anybody seeing this: You are not meant to try to correct this by hand-editing yourself. The tool is not built as a linter, but as a generator. The tool checks for byte-by-byte equality for what it generates.

PR #121, opened 2022-09-27
Make table generation conflicts easier to debug (with colors!)

hughesjj:
[Related issue](https://github.com/open-telemetry/build-tools/issues/120)

# Context
Produces a visual difference in the output to make an error message involved in table generation more contextually meaningful and easier to debug.

I use ANSI color escape codes that [should be compatible](https://en.wikipedia.org/wiki/ANSI_escape_code#Platform_support) between any "modern" terminal in windows, mac, or linux.  The below screenshots were taken on linux using `wezterm`.  I've done my best to make the mechanism for coloring diffs maintainable and understandable. []()

Currently, I output the entire difflib output.  For enhancements on large markdown files, we could consider replacing the "equals" opcode case with some "relative offset" information.  However that's complicated and likely warrants further discussion, so for now I propose we display the full output.  The full output is supposed to be human readable anyway, so it shouldn't be *too* long in practice.

## Current error message
image

## Error message after this review
image


semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
```diff
@@ -37,6 +39,73 @@
 from opentelemetry.semconv.templating.markdown.options import MarkdownOptions
 
 
+class VisualDiffer:
```
joaopgrassi, line 42: I'm not sure how I feel about this new class being "together" with the rest of the actual functionality of the generator. I wonder if it wouldn't be better if it was in a separated file for ex.

Oberon00:
Please give the PR a more descriptive / concrete title, like "Add diff output on errors"

semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
joaopgrassi: I guess just a file on its own, TableDiffHelper or something :D

semantic-conventions/src/opentelemetry/semconv/templating/markdown/utils.py
```diff
@@ -0,0 +1,88 @@
...
+        Prints git-like colored diff using ANSI terminal coloring.
+        Diff is "from a to b", that is, red text is text deleted in `a`
+        while green text is new to `b`
+        """
+        if not os.environ.get("COLORED_DIFF"):
+            # return "\n".join(difflib.context_diff(a, b))
+            added = lambda text: f"added:'{text}'"
+            removed = lambda text: f"removed:'{text}'"
+        else:
+            added = lambda text: cls.colorize_text(0, 255, 0, text)
```
jsuereth, line 67: RGB is 6-bit not 8-bit, so I think any value > 64 doesn't matter.

semantic-conventions/src/tests/data/markdown/table_generation_conflict/expected-no-colors.md
```diff
@@ -0,0 +1,15 @@
...
+
+**Status**: [Experimental](../../../document-status.md)
+
+**type:** `foo.bar`
+
+**Description:** Test case
+
+<!-- semconv redis -->
+
+| Atadded:'t'ributeremoved:' ' | Type   | Description added:' '| Examples | Requirement Level |
```
jsuereth, line 11: Am I missing something or are these `added:'t'` unexpected.

jsuereth (commented):
Overall love this nice addition, but not sure I understand how the tests are functioning given the examples.

Also, as a side note, when I last did ANSI color, I stuck with the 255 index table using something like this: https://github.com/jsuereth/streamerz/blob/master/ansi/src/main/scala/com/jsuereth/ansi/Ansi.scala#L64

jsuereth (approved):

semantic-conventions/src/tests/data/markdown/table_generation_conflict/expected-no-colors.md
jsuereth: This looks a bit more readable to me.

joaopgrassi:
@hughesjj this is been open for much long.. would you have the time to fix the conflicts so we can merge this? I think it's a great addition.

joaopgrassi (approved):

PR #79, opened 2022-01-12
Metrics table generation from yaml

jamesmoessis:
This PR adds functionality such that metrics can be parsed from yaml and relevant information generated in markdown. 

~~This can regenerate the structure of the [http metric semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md) that we have today, including the "Attribute alternatives" which can be controlled via the `constraints` in the yaml.~~

There are two types of tables that must be generated from `MetricSemanticConvention`: 
- attributes (same as before)
- metrics (id, instrument, units)

`<!-- semconv metric.http.client -->` - creates an attribute table as normal.

`<!-- semconv metric.http.client(metric_table) -->` - creates a metric table of all `metrics` that are under `metric.http.client` in the yaml file. If you have other ideas about how the yaml file should be structured, let me know (see `http_metrics.yaml`).

The `metric` prefix ID is used to prevent clash with the tracing semconvs of the same ID. 

It's quite few lines of actual code, most of this PR is testing. 

cc @ocelotl @jsuereth

PR #131, opened 2023-01-20
Move footnotes back together with the rendered table in Markdown

arminru:
When trying to integrate the new tooling for metrics with the spec repo I noticed that #79 accidentally introduced a breaking change in markdown rendering.
It would move the footnotes belonging to attribute tables apart from them. https://github.com/open-telemetry/build-tools/releases/tag/v0.15.0 first prints the list defining sampling relevant attributes and only then prints the footnotes.

### DIFF
--- a/semantic-conventions/CHANGELOG.md
+++ b/semantic-conventions/CHANGELOG.md
@@ -4,6 +4,11 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
+## v0.15.1
+
+- Move footnotes back together with the rendered table in Markdown
+  ([#131](https://github.com/open-telemetry/build-tools/pull/131))
+
 ## v0.15.0
 
 - Add a semantic convention type for Metrics ("metric" and "metric_group") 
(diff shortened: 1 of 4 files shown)


joaopgrassi (approved):

semantic-conventions/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
+## v0.15.1
+
+- Move footnotes back together with the rendered table in Markdown
```
Oberon00, line 9: This is a bit confusing since from the test it looks like they were completely gone. But I see that you also changed the test YAML. So the footnotes would have moved further away from the table? I think it's no big deal either way, but since unnecessary changes add noise in the spec repo, this reversion to the old behavior is good.

All in all, I still would love to see somebody tackle https://github.com/open-telemetry/opentelemetry-specification/issues/1925 to entirely get rid of this splintered output.

Oberon00 (approved):

semantic-conventions/CHANGELOG.md
arminru: Yes, it just moved it around and there was no test covering that before but they were not lost.

+1 on improving the output overall at some point 🙂

PR #135, opened 2023-02-17
Rename Optional attribute requirement level to Opt-In

trask:
Based on suggestion by @arminru and others.

Needed before we mark [attribute-requirement-level.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md) as stable (which is needed before we mark HTTP semantic conventions stable, see https://github.com/open-telemetry/opentelemetry-specification/issues/3219)

### DIFF
--- a/semantic-conventions/semconv.schema.json
+++ b/semantic-conventions/semconv.schema.json
@@ -341,7 +341,7 @@
 				{
 					"properties": {
 						"requirement_level": {
-							"description": "specifies the attribute requirement level. Can be 'required', 'conditionally_required', 'recommended', or 'optional'. When omitted, the attribute is 'recommended'. When set to 'conditionally_required', the string provided as <condition> MUST specify the conditions under which the attribute is required.",
+							"description": "specifies the attribute requirement level. Can be 'required', 'conditionally_required', 'recommended', or 'opt_in'. When omitted, the attribute is 'recommended'. When set to 'conditionally_required', the string provided as <condition> MUST specify the conditions under which the attribute is required.",
 							"oneOf": [
 								{
 									"const": "required"
@@ -378,7 +378,7 @@
 									]
 								},
 								{
-									"const": "optional"
+									"const": "opt_in"
 								}
 							]
 						},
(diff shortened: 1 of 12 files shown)


yurishkuro (approved):
+1

yurishkuro:
I like the name change, but did we do an analysis that all usages of the "optional" level are actually opt-in?  Because the two are not exactly equivalent:
  * optional is something that instrumentation may or may not support (fairly similar to recommended, but to lesser degree)
  * opt-in is something that instrumentation MUST require an active opt-in before enabling the attribute

Is it possible that we need to keep both concepts and just do a once-over review if some currently optional attrs should really be opt-in? if we only need opt-in (given that optional ~= recommended), then I don't think this is how many people may actually understood optional previously, and as a result the attrs may be proactively emitted by instrumentation without asking user to enable them.

trask:
I just sent a draft spec PR to explore that: https://github.com/open-telemetry/opentelemetry-specification/pull/3228

Luckily, `optional` isn't used in many places yet, so only two I'm unsure about which meaning was intended, I left comments to try to get clarification on those two places.

Oberon00:
>  Because the two are not exactly equivalent

I think there is a definition of optional that says it's "opt-in": https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-requirement-level.md#optional

> ## Optional
>
> Instrumentations SHOULD populate the attribute if and only if the user configures the instrumentation to do so. Instrumentation that doesn't support configuration MUST NOT populate Optional attributes.

trask:
> Maybe it would be better to just change the markdown output, as everything else is a breaking change.

I split out the markdown rendering into a separate PR #137 in case there's a preference to move forward with that first and then we can discuss the breaking change on its own.

Oberon00 (approved):
I approved both PRs, depending on how you want to handle it in the spec and potentially code generators (I doubt anybody actually uses the requirement levels there though, except maybe for printing them out).

arminru (approved):

trask:
> I approved both PRs, depending on how you want to handle it in the spec and potentially code generators (I doubt anybody actually uses the requirement levels there though, except maybe for printing them out).

thx! I'd prefer to merge this PR and trigger a new release of build-tools, then I will proceed with https://github.com/open-telemetry/opentelemetry-specification/pull/3228

trask:
are there enough approvals on https://github.com/open-telemetry/opentelemetry-specification/pull/3228 to merge this? thx!

arminru:
@trask I believe so 🙂
If any strong disagreement comes up still we could revert build-tools before pulling it into the spec.

Issue #122, opened 2022-10-12
Semconv code generator example does not work for opentelemetry-specifications 1.14.0


marcalff:
Repo opentelemetry-cpp successfully used the semconv code generator already on 1.13.0 specs.
See the generate script used:

[generate.sh](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/buildscripts/semantic-convention/generate.sh)

Now trying to upgrade to 1.14.0 opentelemetry-specification:

`SEMCONV_VERSION=1.14.0`

The build fails with this error message:

```
opentelemetry.semconv.model.exceptions.ValidationError: Semantic Convention trace-exception reference `exception.type` but it cannot be found! - @2:5
```

In the 1.14.0 specs, indeed, file
`semantic_conventions/trace/trace-exception.yaml`
uses the exception defined in 
`semantic_conventions/exception.yaml`

Because the docker mount point used is `semantic_conventions/trace/`,
the docker image does not see file `exceptions.yaml`

```
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions/trace:/source \
  ...
```

I hope this helps to narrow down the root cause.
No idea about a fix however.

Please fix and adjust the example to work with the latest semantic conventions from the specs,
as this is needed for each language SIG in general.

[resolved] Blocking for open-telemetry/opentelemetry-cpp#1671

Oberon00:
Have you tried just running it over everything, i.e. `-v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions`? without `/trace`?
If you only want span conventions, I think you can use the `--only span` (https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/opentelemetry/semconv/main.py#L211-L215)

marcalff:
Thanks @Oberon00 

Adjusted the generation script to:

```
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/api/include/opentelemetry/trace/:/output \
  otel/semconvgen:$GENERATOR_VERSION \
  --only span \
  -f /source code \
  --template /templates/SemanticAttributes.h.j2 \
  --output /output/semantic_conventions.h \
  -Dsemconv=trace \
  -Dclass=SemanticConventions \
  -DschemaUrl=$SCHEMA_URL \
[...]

PR #156, opened 2023-03-09
Support reading additional files with references

### LINKED ISSUE #155: Support referencing attributes across signals in codegen
See https://github.com/open-telemetry/opentelemetry-specification/issues/3299.
E.g. `user_agent.original` is defined within `attribute_group` in opentelemetry-specification/semantic_conventions folder and referenced in resource semconv here https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/resource/browser.yaml

Code generator [allows](https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/opentelemetry/semconv/main.py) to specify `source` folder and would generate all semconv from there, applying `--only` or `--exclude` flags to the source.  It uses the same subset of files to resolve attributes and generate code for them.

When browser resource semconv is generated, we need to use all resource + cross-signal attributes to resolve attributes, but generate only those that  are in resource. there is no such option today.

lmolkova:
Fixes #155

Adds `--references-only` flag to `semconvgen` which allows to specify additional sources for referenced (extended, included) semconv.

E.g. browser resource semconv references general `user_agent.original` attribute. When code for resources is generated, `user_agent.original` needs to be resolved, even though `user_agent` semconv is not generated.

Usage (tested on Java):

[All semconv except resources](https://github.com/open-telemetry/opentelemetry-java/blob/328dcf075a827ebe710274d020baafb978091b01/buildscripts/semantic-convention/generate.sh#L24-36)

```diff
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/:/output \
  otel/semconvgen:$GENERATOR_VERSION \
-  --exclude resource/** \
+  --reference-only resource/** \
  -f /source code \
 ...
```
 New flag excludes semconvs from the output, but allows references to them.

Resources:

```diff
docker run --rm \
- -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions/resource/:/source \
+ -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/semconv/src/main/java/io/opentelemetry/semconv/resource/attributes/:/output \
  semconvgen:$GENERATOR_VERSION \
+ --only resources \
+ --reference-only *.yaml \
  -f /source code \
....
```

I.e. from all semconv output resource conventions, and use yaml files from root folder `semantic_conventions` to resolve references.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/main.py
+++ b/semantic-conventions/src/opentelemetry/semconv/main.py
@@ -32,16 +32,24 @@ def parse_semconv(args, parser) -> SemanticConventionSet:
     semconv = SemanticConventionSet(args.debug)
     find_yaml(args)
     for file in sorted(args.files):
-        if not file.endswith(".yaml") and not file.endswith(".yml"):
+        if not is_yaml_file(file):
             parser.error(f"{file} is not a yaml file.")
         semconv.parse(file)
+
+    if args.reference_only is not None:
+        reference_files = [f for f in glob_file_list(args.yaml_root, args.reference_only) if is_yaml_file(f)]
+        for file in sorted(reference_files):
+            semconv.parse(file, True)
+
     semconv.finish()
     if semconv.has_error():
         sys.exit(1)
     return semconv
 
+def is_yaml_file(file: str) -> bool:
+    return file.endswith(".yaml") or file.endswith(".yml")
 
-def exclude_file_list(folder: str, pattern: str) -> List[str]:
+def glob_file_list(folder: str, pattern: str) -> List[str]:
     if not pattern:
         return []
     sep = "/"
@@ -85,7 +93,7 @@ def process_markdown(semconv, args):
         enable_deprecated=args.md_enable_deprecated,
         use_badge=args.md_use_badges,
         break_count=args.md_break_conditional,
-        exclude_files=exclude_file_list(args.markdown_root, args.exclude),
+        exclude_files=glob_file_list(args.markdown_root, args.exclude),
     )
     md_renderer = MarkdownRenderer(args.markdown_root, semconv, options)
     md_renderer.render_md()
@@ -94,8 +102,13 @@ def process_markdown(semconv, args):
 def find_yaml(args):
     if args.yaml_root is not None:
         exclude = set(
-            exclude_file_list(args.yaml_root if args.yaml_root else "", args.exclude)
+            glob_file_list(args.yaml_root if args.yaml_root else "", args.exclude)
         )
+        reference_only = set(
+            glob_file_list(args.yaml_root if args.yaml_root else "", args.reference_only)
+        )
+        exclude.update(reference_only) 
+
         yaml_files = set(
             glob.glob(f"{args.yaml_root}/**/*.yaml", recursive=True)
         ).union(set(glob.glob(f"{args.yaml_root}/**/*.yml", recursive=True)))
@@ -222,6 +235,9 @@ def setup_parser():
     )
     parser.add_argument(
         "--exclude", "-e", help="Exclude the matching files using GLOB syntax", type=str
+    )
+    parser.add_argument(
+        "--reference-only", "-r", help="Additional files to resolve references only (and extends) using GLOB syntax", type=str
     )
     parser.add_argument(
         "files",
--- a/semantic-conventions/src/opentelemetry/semconv/main.py
+++ b/semantic-conventions/src/opentelemetry/semconv/main.py
@@ -37,7 +37,11 @@ def parse_semconv(args, parser) -> SemanticConventionSet:
         semconv.parse(file)
 
     if args.reference_only is not None:
-        reference_files = [f for f in glob_file_list(args.yaml_root, args.reference_only) if is_yaml_file(f)]
+        reference_files = [
+            f
+            for f in glob_file_list(args.yaml_root, args.reference_only)
+            if is_yaml_file(f)
+        ]
         for file in sorted(reference_files):
             semconv.parse(file, True)
 
@@ -46,8 +50,10 @@ def parse_semconv(args, parser) -> SemanticConventionSet:
         sys.exit(1)
     return semconv
 
+
 def is_yaml_file(file: str) -> bool:
     return file.endswith(".yaml") or file.endswith(".yml")
+
 
 def glob_file_list(folder: str, pattern: str) -> List[str]:
     if not pattern:
@@ -105,9 +111,11 @@ def find_yaml(args):
             glob_file_list(args.yaml_root if args.yaml_root else "", args.exclude)
         )
         reference_only = set(
-            glob_file_list(args.yaml_root if args.yaml_root else "", args.reference_only)
+            glob_file_list(
+                args.yaml_root if args.yaml_root else "", args.reference_only
+            )
         )
-        exclude.update(reference_only) 
+        exclude.update(reference_only)
 
         yaml_files = set(
             glob.glob(f"{args.yaml_root}/**/*.yaml", recursive=True)
@@ -237,7 +245,10 @@ def setup_parser():
         "--exclude", "-e", help="Exclude the matching files using GLOB syntax", type=str
     )
     parser.add_argument(
-        "--reference-only", "-r", help="Additional files to resolve references only (and extends) using GLOB syntax", type=str
+        "--reference-only",
+        "-r",
+        help="Additional files to resolve references only (and extends) using GLOB syntax",
+        type=str,
     )
     parser.add_argument(
         "files",
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -300,9 +300,10 @@ class SemanticConventionSet:
 
     debug: bool
     models: typing.Dict[str, BaseSemanticConvention] = field(default_factory=dict)
+    extended_models: typing.Dict[str, BaseSemanticConvention] = field(default_factory=dict)
     errors: bool = False
 
-    def parse(self, file):
+    def parse(self, file, ref_only=False):
         with open(file, "r", encoding="utf-8") as yaml_file:
             try:
                 semconv_models = parse_semantic_convention_groups(yaml_file)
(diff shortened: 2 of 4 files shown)


lmolkova:
@Oberon00 @arminru can you please take a look? Apparently codegen is broken for 1.19.0 release (https://github.com/open-telemetry/opentelemetry-specification/issues/3299).

semantic-conventions/src/opentelemetry/semconv/templating/code.py
```diff
@@ -244,6 +244,7 @@ def render(
                 template.stream(data).dump(output_name)
         else:
             data = self.get_data_single_file(semconvset, template_path)
+            print (data)
```
Oberon00, line 247: Leftover?

semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -223,6 +244,12 @@ def setup_parser():
     parser.add_argument(
         "--exclude", "-e", help="Exclude the matching files using GLOB syntax", type=str
     )
+    parser.add_argument(
+        "--reference-only",
+        "-r",
+        help="Additional files to resolve references only (and extends) using GLOB syntax",
```
Oberon00, line 250: I think this feature warrants a bit more description. This suggestion is also a check if I understood this PR correctly 😃 

```suggestion
        help="""Additional files that will be loaded and available for referencing, but will not be part of the output.
        
        As semantic conventions can, in various ways, e.g. using `ref` or `include`, reference other conventions you
        sometimes need to add files/groups to your input but you don't want the groups in there be processed by
        the code generator template or markdown generator. In this case, add only the subset of files you want to
        process with your template / markdown generator to `--yaml-root`, and add any additional required
        (transitively depended-upon) files with `--reference-only`.
        """,
```

semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -223,6 +244,12 @@ def setup_parser():
     parser.add_argument(
         "--exclude", "-e", help="Exclude the matching files using GLOB syntax", type=str
     )
+    parser.add_argument(
+        "--reference-only",
+        "-r",
```
Oberon00, line 249: I wonder if we actually want a short-name for this flag.

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -300,9 +300,12 @@ class SemanticConventionSet:
 
     debug: bool
     models: typing.Dict[str, BaseSemanticConvention] = field(default_factory=dict)
+    extended_models: typing.Dict[str, BaseSemanticConvention] = field(
+        default_factory=dict
+    )
     errors: bool = False
 
-    def parse(self, file):
+    def parse(self, file, ref_only=False):
```
Oberon00, line 308: Maybe instead add `parse_into(self, file, models)`?

If you prefer keeping the boolean argument, please add `*` before to enforce using named arguments for it:

```suggestion
    def parse(self, file, *, ref_only=False):
```

semantic-conventions/src/tests/semconv/model/test_correct_parse.py
```diff
@@ -727,6 +727,68 @@ def test_scope_attribute(self):
         }
         self.semantic_convention_check(list(semconv.models.values())[0], expected)
 
+    def test_reference_only_ref(self):
+        semconv = SemanticConventionSet(debug=False)
+        semconv.parse(self.load_file("yaml/http.yaml"))
+        self.assertEqual(len(semconv.models), 3)
+        semconv.parse(self.load_file("yaml/general.yaml"), True)
```
Oberon00, line 734:
```suggestion
        semconv.parse(self.load_file("yaml/general.yaml"), ref_only=True)
```

etc
(or see comment before for maybe using an API that does not rely on a boolean)

Oberon00 (approved):

Oberon00:
> As in, if I give a path to where the yaml schema files are to -f and that includes all the necessary references within it 

I think this is already the case. IIUC, this PR is actually about adding a way to exclude files from output generation but not from internal model construction (see https://github.com/open-telemetry/build-tools/pull/156#discussion_r1130663829).

`--exclude` I think completely excludes the files from being loaded. `--only` should be applied after parsing I believe.

I wonder if this PR is maybe going into the wrong direction, as it adds more reliance on the directory structure. Maybe globbing on the group ID would be better. Or maybe `--only` would already solve the problem, or if not, could be adjusted a bit?

Oberon00:
Actually I now looked at how `--only` is implemented, see https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/src/opentelemetry/semconv/main.py#L68 and called function. I think that approach would also work for this PR and would be much simpler. At the moment, the source file path of a convention is not part of the model, but it could be added, or alternatively, the approach could be changed to filter by group ID (regex/glob) instead of filepath.

Oberon00:
Another option would be: instead of adding additional paths, you would add the actual root as --yaml-root, and you would add another option `--only-paths` that then restricts the output. So instead of `--yaml-root some-subdirectory-I-want-to-generate-output-for --reference-only common-directory` (which breaks as soon as there is one than more common directory, or would become cumbersome), you would do `--yaml-root actual-toplevel-yaml-root --only-paths some-subdirectory-I-want-to-generate-output-for`

marcalff:
Thanks for looking into this.

I don't have specific comments about the implementation itself, and which options to pass to generate code.

From an end-user point of view, and based on my (very limited) understanding of semconv code generation, expectations are as follows:

* (1) Example documentation that works, see issue https://github.com/open-telemetry/build-tools/issues/122
* (2) Code generation separate for traces / resources / metrics (when available), so that semconv per signal can go into different files, which contains only the applicable constants (for C++, trace semconv in trace.h, resource semconv in resource.h, etc), hence the use of `-only span`, ...
* (3) A stable docker run command, that does not change every spec release.

A docker command like:

```
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  --only span \
 ...
```

seems to meet these expectations, assuming it can be supported.

If there is a need to:

* add more -v mount options to add some *.yaml files
* add more --exclude options to hide some other *.yaml files

every time the spec releases a different set of *.yaml files, due to changes in internal dependencies, this becomes more difficult to deal with, and is more subject to errors.

Changing the docker run command once this is resolved is not an issue, but having to change it all the time (3) is a concern here.

If somehow these expectations are not realistic, please clarify what we (users in languages) are supposed to do then (see (1)).

Regards.

Oberon00:
@marcalff

Thanks for your comment. I was of the opinion that the docker command you mention should work even today. If not, I'd like to know what the problem is.

lmolkova:
to address @marcalff concern, I think
- we need to add a sample codegen dry run to -spec repo workflow so that authors of PR should see if the change breaks codegen
- we should reasonably unify how codegen is used (I see significant differences in Java and c++)

It's also the first time for me looking into codegen and here I'd really like to reasonably unblock ourselves for 1.19.0 release and take steps to prevent it. I wonder if these steps are good enough.

Regarding reusing current functionality without a new flag and allowing multiple `--only` specs:
I don't think it's a good idea to allow referring to metrics attributes to refer span attributes and vice versa. We can solve it in several ways:
1. explicit inclusion of files I added here gives the most of control
2. hardcoding what's allowed in build-tools (i.e. traces can refer to `resource` and `attribute_group` attributes, but not `metric` ones)

The opt2 is somewhat controversial and I'm not ready to drive it forward.

Oberon00:
> Regarding reusing current functionality without a new flag and allowing multiple --only specs: I don't think it's a good idea to allow referring to metrics attributes from spans and vice versa.

I think that has very little to do with the issue here. I think this should be either enforced by the tool (option 2) or by spec reviewers. Option 1 would be "enforcement by hard UX"

tsloughter:
> At the moment, the source file path of a convention is not part of the model

Right, I think that is what I was trying to suggest.

I'm trying to catch up. I don't think having multiple values to `--only` will work because of how the script is used to generate a specific file. Like in Go:

```
$(SEMCONVGEN) -i "$(OTEL_SPEC_REPO)/semantic_conventions/." --only=span -p conventionType=trace -f trace.go -t "$(SEMCONVPKG)/template.j2" -s "$(TAG)"
```

And Python:

```
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions/trace:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/opentelemetry-semantic-conventions/src/opentelemetry/semconv/trace/:/output \
  otel/semconvgen:$OTEL_SEMCONV_GEN_IMG_VERSION \
  -f /source code \
  --template /templates/semantic_attributes.j2 \
  --output /output/__init__.py \
  -Dclass=SpanAttributes
```

lmolkova:
> I don't think having multiple values to --only will work because of how the script is used to generate a specific file. Like in Go:

we'll have to update generate scripts anyway and the common pattern can be:

```bash
-v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
--only  span,attribute_group,event
```

this will preserve current attribute locations in files.

an alternative could be to run script per semconv type and move attributes into individual files

```bash
-v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
--only  span
```

```bash
-v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
--only  attribute_group
```

```bash
-v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
--only  event
```

lmolkova:
Did multiple `--only` here - https://github.com/open-telemetry/build-tools/pull/157, tried on Java, works just fine and much simpler.

Oberon00:
Should attribute_group even be required to be mentioned? I think for referencing, you don't need to include it in `--only`

lmolkova:
closing in favor of #157

PR #157, opened 2023-03-10
Allow multiple semconv in --only flag

lmolkova:
Alternative approach to #156

Usage diff: 

Java generates all semconv except resources into one file
```diff
docker run --rm \
  -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/:/output \
  otel/semconvgen:$GENERATOR_VERSION \
-  --exclude resource/** \
+  --only span,event,attribute_group,scope \
  -f /source code \
 ...
```

Resources:

```diff
docker run --rm \
- -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions/resource/:/source \
+ -v ${SCRIPT_DIR}/opentelemetry-specification/semantic_conventions:/source \
  -v ${SCRIPT_DIR}/templates:/templates \
  -v ${ROOT_DIR}/semconv/src/main/java/io/opentelemetry/semconv/resource/attributes/:/output \
  semconvgen:$GENERATOR_VERSION \
+ --only resource \
  -f /source code \
....
```

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/main.py
+++ b/semantic-conventions/src/opentelemetry/semconv/main.py
@@ -59,12 +59,13 @@ def filter_semconv(semconv, types):
             if model.GROUP_TYPE_NAME in types
         }
 
+
 def main():
     parser = setup_parser()
     args = parser.parse_args()
     check_args(args, parser)
-    semconv_filter = parse_only_filter(args, parser)
     semconv = parse_semconv(args, parser)
+    semconv_filter = parse_only_filter(args, parser)
     filter_semconv(semconv, semconv_filter)
     if len(semconv.models) == 0:
         parser.error("No semantic convention model found!")
@@ -113,10 +114,12 @@ def parse_only_filter(arguments, parser):
     if not arguments.only:
         return None
 
-    types = [t.strip() for t in arguments.only.split(",")]        
+    types = [t.strip() for t in arguments.only.split(",")]
     unknown_types = [t for t in types if t not in CONVENTION_CLS_BY_GROUP_TYPE.keys()]
     if unknown_types:
-        parser.error(f"Unknown semconv names in `--only` option: '{', '.join(unknown_types)}'")
+        parser.error(
+            f"Unknown semconv names in `--only` option: '{', '.join(unknown_types)}'"
+        )
         sys.exit(1)
     return types
 
@@ -220,7 +223,6 @@ def setup_parser():
     parser.add_argument(
         "--debug", "-d", help="Enable debug output", action="store_true"
     )
-
     parser.add_argument(
         "--only",
         type=str,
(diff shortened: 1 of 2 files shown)


semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -211,7 +211,9 @@ def setup_parser():
     parser.add_argument(
         "--only",
         choices=list(CONVENTION_CLS_BY_GROUP_TYPE.keys()),
-        help="Process only semantic conventions of the specified type.",
+        type=str,
+        nargs="+",
```
Oberon00, line 215: I think `+` is usually a bad idea since it may eat too many arguments. Maybe a manual split by `,` would be better, or the same method we use for codegeration's `-D`, method=append https://github.com/lmolkova/build-tools/blob/alt-fix-codegen-119/semantic-conventions/src/opentelemetry/semconv/main.py#L139. But my argparse skills are a bit rusty, so not sure maybe it is OK.

lmolkova: Did comma-separated string.  `--only span --only event` look weird and I'd need to find a better alias.

semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -51,12 +51,13 @@ def exclude_file_list(folder: str, pattern: str) -> List[str]:
     return file_names
 
 
-def filter_semconv(semconv, type_filter):
-    if type_filter:
+def filter_semconv(semconv, types_filter):
+    if types_filter:
+        types = types_filter.replace(" ", "").split(",")
```
Oberon00, line 56: What is the goal of replacing spaces here? Maybe you wanted to do this?

```suggestion
        types = [t.strip() for t in types_filter.split(",")]
```

Oberon00 (approved):

semantic-conventions/src/opentelemetry/semconv/main.py
Oberon00, line 56: It could also make sense to validate that all the group type names are valid (known). Otherwise, typos in group type names here will lead to the group being skipped

lmolkova: sure, done

lmolkova:
@Oberon00 thanks for the review! 

@arminru could you please take a look and merge if you're fine with this change?

Oberon00:
@lmolkova I guess this one then replaces #156? Maybe add "Closes #156" to the PR description to auto-close it on merge & have it linked in the sidebar.

lmolkova:
@open-telemetry/specs-approvers can someone please take a look and merge? it unblocks code generation for 1.19.0 release https://github.com/open-telemetry/opentelemetry-specification/issues/3299

PR #170, opened 2023-04-25
Add frozen stability level and change default stability level to experimental

lmolkova:
In preparation for https://github.com/open-telemetry/opentelemetry-specification/issues/3279, it'd be great to have a formal way to declare attributes as frozen.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/main.py
+++ b/semantic-conventions/src/opentelemetry/semconv/main.py
@@ -82,6 +82,7 @@ def process_markdown(semconv, args):
     options = MarkdownOptions(
         check_only=args.md_check,
         enable_stable=args.md_stable,
+        enable_frozen=args.md_frozen,        
         enable_experimental=args.md_experimental,
         enable_deprecated=args.md_enable_deprecated,
         use_badge=args.md_use_badges,
@@ -200,6 +201,12 @@ def add_md_parser(subparsers):
         required=False,
         action="store_true",
     )
+    parser.add_argument(
+        "--md-frozen",
+        help="Add labels to attributes marked as frozen.",
+        required=False,
+        action="store_true",
+    )    
     parser.add_argument(
         "--md-experimental",
         help="Add labels to attributes marked as experimental.",
--- a/semantic-conventions/src/opentelemetry/semconv/main.py
+++ b/semantic-conventions/src/opentelemetry/semconv/main.py
@@ -82,7 +82,7 @@ def process_markdown(semconv, args):
     options = MarkdownOptions(
         check_only=args.md_check,
         enable_stable=args.md_stable,
-        enable_frozen=args.md_frozen,        
+        enable_frozen=args.md_frozen,
         enable_experimental=args.md_experimental,
         enable_deprecated=args.md_enable_deprecated,
         use_badge=args.md_use_badges,
@@ -206,7 +206,7 @@ def add_md_parser(subparsers):
         help="Add labels to attributes marked as frozen.",
         required=False,
         action="store_true",
-    )    
+    )
     parser.add_argument(
         "--md-experimental",
         help="Add labels to attributes marked as experimental.",
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -39,6 +39,7 @@ class StabilityLevel(Enum):
     STABLE = 1
     EXPERIMENTAL = 2
     DEPRECATED = 3
+    FROZEN = 4
 
 
 def unique_attributes(attributes):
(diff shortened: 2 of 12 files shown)


trask (approved):

lmolkova:
@Oberon00 @arminru could you please take a look? Thanks!

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -192,7 +193,7 @@ def parse(
                 )
                 msg = f"Semantic convention stability set to deprecated but attribute '{attr_id}' is {stability}"
                 raise ValidationError.from_yaml_pos(position, msg)
-            stability = stability or semconv_stability or StabilityLevel.STABLE
+            stability = stability or semconv_stability or StabilityLevel.EXPERIMENTAL
```
Oberon00, line 196: This seems like a good change, but also important enough to include in the PR title

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -314,6 +315,7 @@ def check_stability(stability_value, position):
             "deprecated": StabilityLevel.DEPRECATED,
             "experimental": StabilityLevel.EXPERIMENTAL,
             "stable": StabilityLevel.STABLE,
+            "frozen": StabilityLevel.FROZEN,
```
Oberon00, line 318: "Frozen" is not a stability level AFAIK but orthogonal to it. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md#feature-freeze

Oberon00 (changes requested):

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
lmolkova: trying to understand the concern: is it a different term or you'd rather see it as a separate flag?

1. If it's a term,  we chatted with @reyang and @trask  about `frozen` vs `feature-freeze` and considered (in the long term) changing the `feature-freeze` to `frozen` to apply better to things like attributes and be an adjective following `stable` or `deprecated` pattern.  Go use [`Frozen`](https://zchee.github.io/golang-wiki/Frozen/) convention for packages to describe that no new features are accepted.
2. If you'd rather see it as a separate flag: `feature-freeze` is still a valid document status used interchangeably with `Stable` or `Experimental`. Having two attribute parameters would be confusing, as I can write something meaningless as
  ```yaml
  stability: stable | deprecated
  feature-freeze: false
  ```

semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -200,6 +201,12 @@ def add_md_parser(subparsers):
         required=False,
         action="store_true",
     )
+    parser.add_argument(
```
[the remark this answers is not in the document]
lmolkova: Do you mean that when an attribute becomes frozen, we change it's brief to mention it there? It's a viable solution, but then it's  harder to read and review, harder to follow the same pattern, and easier to overlook.
It's also not possible to add any checks (if we want to).

I also think it could be useful to add a stability column and represent stability information separately from brief.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
Oberon00: 2 is what I mean. As you can see in the linked document, a spec document has a lifecylce status and it can be feature-frozen or not. So indeed:

```yaml
stability: stable | deprecated | experimental(default)
feature-freeze: true | false(default)
```

At least I would suggest adopting the same system for whole spec documents and semantic convention elements, otherwise it could get confusing.

I have no strong opinion on whether that is ideal or if "frozen" should actually be a lifecylce status (optional step between experimental and stable?), but then I think it should be changed in the specs document-status spec first (which itself doesn't state a document status which would make it experimental, whether intentional or not)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md

trask: https://github.com/open-telemetry/opentelemetry-specification/pull/3456

dyladan:
I was updating the JS semconv package and was very surprised to see the default stability level is stable. We're trying to publish stable separately from experimental in preparation for semconv like HTTP becoming stable soon. Is there any way we can move this forward or at least change the default to experimental?

arminru:
@dyladan I carved the default part out into https://github.com/open-telemetry/build-tools/pull/189 since the discussions on https://github.com/open-telemetry/opentelemetry-specification/pull/3456 might still take some time to resolve.

lmolkova:
Closing this until we have conclusion on https://github.com/open-telemetry/opentelemetry-specification/issues/3459 and https://github.com/open-telemetry/oteps/pull/232

Issue #15, opened 2020-10-19
"brief"s and "note"s of groups should be (optionally?) included in Markdown output


Oberon00:
We allow the whole group of semantic conventions to have a brief and note, but it is not possible yet to include that in the markdown output.

(from https://github.com/open-telemetry/opentelemetry-specification/pull/1054#discussion_r505390006)

PR #177, opened 2023-05-31
Render notes on metric semconv tables

### LINKED ISSUE #167: Add note for metrics yaml definitions
This isn't a blocker, as we can add the note manually to the markdown file for now, but it would be nice to add the note in the yaml definition.

Related to https://github.com/open-telemetry/opentelemetry-specification/issues/3418 and https://github.com/open-telemetry/opentelemetry-specification/pull/3352#issuecomment-1517987640

joaopgrassi:
resolves #167, part of #15.

This only targets metrics, as for that we have a well-defined way to render in markdown (`to_markdown_metric_table`). For other convention types such as `resource` we don't have a well-defined markdown "template" on how we want them to look so I left it for now. 

cc @trask

### DIFF
--- a/semantic-conventions/src/tests/data/yaml/metrics.yaml
+++ b/semantic-conventions/src/tests/data/yaml/metrics.yaml
@@ -10,6 +10,8 @@ groups:
         type: string
         brief: 'Type of egg.'
         examples: ["chicken", "emu", "dragon"]
+        note: >
+          Some notes on attribute
 
   - id: metric.foo.size
     prefix: foo
@@ -18,6 +20,8 @@ groups:
     brief: "Measures the size of foo."
     instrument: histogram
     unit: "{bars}"
+    note: >
+      Some notes on metric.foo.size
     attributes:
       - ref: http.method
         requirement_level: required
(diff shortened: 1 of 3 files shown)


semantic-conventions/src/tests/data/yaml/metrics.yaml
```diff
@@ -18,6 +20,8 @@ groups:
     brief: "Measures the size of foo."
     instrument: histogram
     unit: "{bars}"
+    note: >
```
Oberon00, line 23: Now all the metrics have notes. We should also still test the case without a note.

Oberon00 (approved):

lmolkova (approved):

semantic-conventions/src/tests/data/yaml/metrics.yaml
joaopgrassi: Good point. Fixed in [4f3fb3e](https://github.com/open-telemetry/build-tools/pull/177/commits/4f3fb3e8c148fe04e0a410b857dc5734893bd8ba)

jsuereth (approved):

arminru (approved):

PR #179, opened 2023-06-02
Add semantic convention approvers to CODEOWNERS

arminru:
Quite often changes to our semantic conventions tooling originate from requests and discussions in https://github.com/open-telemetry/semantic-conventions, so I think it would make sense if @open-telemetry/specs-semconv-approvers would also co-own the tooling here. They usually have most of the context and are the front-line stakeholders in semconv tooling.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,4 +16,4 @@
 *   @open-telemetry/specs-approvers
 
 # Owners for the semantic conventions generator tool
-/semantic-conventions/      @open-telemetry/specs-approvers @Oberon00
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -16,4 +16,4 @@
 *   @open-telemetry/specs-approvers
 
 # Owners for the semantic conventions generator tool
-/semantic-conventions/      @open-telemetry/specs-approvers @Oberon00
```
arminru, line 19: Removing @Oberon00's explicit mention since he's already part of @open-telemetry/specs-semconv-approvers 🙂

Oberon00 (approved):

joaopgrassi (approved):

jsuereth (approved):

PR #186, opened 2023-07-07
Enable rendering of "template" attributes in table generation

AlexanderWert:
## Problem
In the semantic conventions we have `template`-style attributes in some places.

Examples are:
- [HTTP request and response headers](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#http-request-and-response-headers): `http.request.header.<key>`, `http.response.header.<key>`
- [Elasticsearch path parts](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md#url-path-parts): `db.elasticsearch.path_parts.<key>` 

This kind of attributes could not be included in the model `.yml` files yet, and thus could not be included in table generation.

## Solution
This PR proposes a fix that enables definition of `template`-style attributes in the YAML files and corresponding table generation.

### How it works
Template-style attributes (i.e. that do not result in a single attribute, but a group of attributes, for example `http.request.header.<key>`) can be defined in a separate `attribute-templates` sub-path in an attribute group.
Attribute templates extend the normal attributes by adding an additional property `parameter_name` that is the name of the parameter in the angle brackets (e.g. `key` in the example above). 

Here is an example YAML file snippet:

```

groups:
  - id: http
    type: span
    prefix: http
    brief: 'This document defines semantic conventions for HTTP client and server Spans.'
    note: >
        These conventions can be used for http and https schemes
        and various HTTP versions like 1.1, 2 and SPDY.
    attribute_templates:                                                 <<---- THIS IS NEW
      - id: request.header
        type: string
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
        parameter_name: key                                             <<---- THIS IS THE NEW PROPERTY
    attributes:
      - id: request.method
        type: string
        requirement_level: required
        sampling_relevant: false
        brief: 'HTTP request method.'
        examples: ["GET", "POST", "HEAD"]
```

This will result in the following markdown result through table generation:

```
<!-- semconv http -->
|---|---|---|---|---|
| `http.request.method` | string | HTTP request method. | `GET`; `POST`; `HEAD` | Required |
| `http.request.header.<key>` | string | HTTP request headers, `<key>` being the normalized HTTP Header name (lowercase, with - characters replaced by _), the value being the header values. | ``http.request.header.content_type=["application/json"]`` | Recommended |
<!-- endsemconv -->
```

Defining such attributes templates in a separate path in the yaml (instead of listing them under the normal attributes) has the following advantages:

1. These attribute templates have slightly different semantic (compared to normal attributes) as they represent a group of attributes (for which the actual name is runtime dynamic / unknown at specification time). 

1. It does not break code generation for attributes, because it's a separate path that is currently just not considered for code generation. Later, we could even build dedicated code generators for template-style attributes for the different languages. That's something that is currently defined 'manually' in all the SDKs.
1. For markdown table generation, this PR already includes template-style attributes as an additional set of attributes. Hence, template-style attributes can be defined in YAML files and are included in table generation.

### DIFF
--- a/semantic-conventions/semconv.schema.json
+++ b/semantic-conventions/semconv.schema.json
@@ -287,7 +287,15 @@
 								"string[]",
 								"int[]",
 								"double[]",
-								"boolean[]"
+								"boolean[]",
+								"template[string]",
+								"template[int]",
+								"template[double]",
+								"template[boolean]",
+								"template[string[]]",
+								"template[int[]]",
+								"template[double[]]",
+								"template[boolean[]]"
 							],
 							"description": "literal denoting the type"
 						},
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -27,7 +27,8 @@ from opentelemetry.semconv.model.utils import (
     validate_values,
 )
 
-TEMPLATE_TYPE_RE = re.compile("template\\[([a-z\\[\\]]+)\\]")
+TEMPLATE_PREFIX = "template["
+TEMPLATE_SUFFIX = "]"
 
 
 class RequirementLevel(Enum):
@@ -77,12 +78,8 @@ class SemanticAttribute:
         return replace(self, inherited=True)
 
     @property
-    def core_type(self):
-        return (
-            AttributeType.get_core_template_type(self.attr_type)
-            if AttributeType.is_template_type(self.attr_type)
-            else self.attr_type
-        )
+    def instantiated_type(self):
+        return AttributeType.get_instantiated_type(self.attr_type)
 
     @property
     def is_local(self):
@@ -382,17 +379,24 @@ class AttributeType:
     def is_template_type(attr_type: str):
         if not isinstance(attr_type, str):
             return False
-        matchObj = TEMPLATE_TYPE_RE.fullmatch(attr_type)
-        if matchObj is not None:
-            return AttributeType.is_simple_type(matchObj.group(1))
-        return False
+
+        return (
+            attr_type.startswith(TEMPLATE_PREFIX)
+            and attr_type.endswith(TEMPLATE_SUFFIX)
+            and AttributeType.is_simple_type(
+                attr_type[len(TEMPLATE_PREFIX) : len(attr_type) - len(TEMPLATE_SUFFIX)]
+            )
+        )
 
     @staticmethod
-    def get_core_template_type(attr_type: str):
-        matchObj = TEMPLATE_TYPE_RE.fullmatch(attr_type)
-        if matchObj is not None:
-            return matchObj.group(1)
-        raise ValidationError(0, 0, f"Cannot retrieve core type from type: {attr_type}")
+    def get_instantiated_type(attr_type: str):
+        if AttributeType.is_template_type(attr_type):
+            return attr_type[
+                len(TEMPLATE_PREFIX) : len(attr_type) - len(TEMPLATE_SUFFIX)
+            ]
+        if AttributeType.is_simple_type(attr_type):
+            return attr_type
+        return "enum"
 
     @staticmethod
     def type_mapper(attr_type: str):
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -27,6 +27,8 @@ from opentelemetry.semconv.model.utils import (
     validate_values,
 )
 
+TEMPLATE_TYPE_RE = re.compile("template\\[([a-z\\[\\]]+)\\]")
+
 
 class RequirementLevel(Enum):
     REQUIRED = 1
@@ -117,7 +119,9 @@ class SemanticAttribute:
                 raise ValidationError.from_yaml_pos(position, msg)
             if attr_id is not None:
                 validate_id(attr_id, position_data["id"])
-                attr_type, brief, examples = SemanticAttribute.parse_id(attribute)
+                attr_type, brief, examples = SemanticAttribute.parse_attribute(
+                    attribute
+                )
                 if prefix:
                     fqn = f"{prefix}.{attr_id}"
                 else:
@@ -231,7 +235,7 @@ class SemanticAttribute:
         return attributes
 
     @staticmethod
-    def parse_id(attribute):
+    def parse_attribute(attribute):
         check_no_missing_keys(attribute, ["type", "brief"])
         attr_val = attribute["type"]
         try:
@@ -242,14 +246,9 @@ class SemanticAttribute:
             position = attribute.lc.data["type"]
             raise ValidationError.from_yaml_pos(position, e.message) from e
         brief = attribute["brief"]
-        zlass = (
-            AttributeType.type_mapper(attr_type)
-            if isinstance(attr_type, str)
-            else "enum"
-        )
-
         examples = attribute.get("examples")
         is_simple_type = AttributeType.is_simple_type(attr_type)
+        is_template_type = AttributeType.is_template_type(attr_type)
         # if we are an array, examples must already be an array
         if (
             is_simple_type
@@ -263,18 +262,31 @@ class SemanticAttribute:
             # TODO: If validation fails later, this will crash when trying to access position data
             # since a list, contrary to a CommentedSeq, does not have position data
             examples = [examples]
-        if is_simple_type and attr_type not in (
-            "boolean",
-            "boolean[]",
-            "int",
-            "int[]",
-            "double",
-            "double[]",
+        if is_template_type or (
+            is_simple_type
+            and attr_type
+            not in (
+                "boolean",
+                "boolean[]",
+                "int",
+                "int[]",
+                "double",
+                "double[]",
+            )
         ):
             if not examples:
                 position = attribute.lc.data[list(attribute)[0]]
                 msg = f"Empty examples for {attr_type} are not allowed"
                 raise ValidationError.from_yaml_pos(position, msg)
+
+        if is_template_type:
+            return attr_type, str(brief), examples
+
+        zlass = (
+            AttributeType.type_mapper(attr_type)
+            if isinstance(attr_type, str)
+            else "enum"
+        )
 
         # TODO: Implement type check for enum examples or forbid them
         if examples is not None and is_simple_type:
@@ -359,6 +371,22 @@ class AttributeType:
         )
 
     @staticmethod
+    def is_template_type(attr_type: str):
+        if not isinstance(attr_type, str):
+            return False
+        matchObj = TEMPLATE_TYPE_RE.fullmatch(attr_type)
+        if matchObj is not None:
+            return AttributeType.is_simple_type(matchObj.group(1))
+        return False
+
+    @staticmethod
+    def get_core_template_type(attr_type: str):
+        matchObj = TEMPLATE_TYPE_RE.fullmatch(attr_type)
+        if matchObj is not None:
+            return matchObj.group(1)
+        raise ValidationError(0, 0, f"Cannot retrieve core type from type: {attr_type}")
+
+    @staticmethod
     def type_mapper(attr_type: str):
         type_mapper = {
             "int": int,
@@ -432,7 +460,9 @@ class EnumAttributeType:
         otherwise it returns the basic type as string.
         """
         if isinstance(attribute_type, str):
-            if AttributeType.is_simple_type(attribute_type):
+            if AttributeType.is_simple_type(
+                attribute_type
+            ) or AttributeType.is_template_type(attribute_type):
                 return attribute_type
             # Wrong type used - raise the exception and fill the missing data in the parent
             raise ValidationError(
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
@@ -77,6 +77,14 @@ class SemanticAttribute:
         return replace(self, inherited=True)
 
     @property
+    def core_type(self):
+        return (
+            AttributeType.get_core_template_type(self.attr_type)
+            if AttributeType.is_template_type(self.attr_type)
+            else self.attr_type
+        )
+
+    @property
     def is_local(self):
         return not self.imported and not self.inherited
 
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -590,6 +590,12 @@ class SemanticConventionSet:
             output.extend(semconv.attributes)
         return output
 
+    def attribute_templates(self):
+        output = []
+        for semconv in self.models.values():
+            output.extend(semconv.attribute_templates)
+        return output
+
 
 CONVENTION_CLS_BY_GROUP_TYPE = {
     cls.GROUP_TYPE_NAME: cls
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -23,6 +23,7 @@ from ruamel.yaml import YAML
 from opentelemetry.semconv.model.constraints import AnyOf, Include, parse_constraints
 from opentelemetry.semconv.model.exceptions import ValidationError
 from opentelemetry.semconv.model.semantic_attribute import (
+    AttributeType,
     RequirementLevel,
     SemanticAttribute,
     unique_attributes,
@@ -111,6 +112,30 @@ class BaseSemanticConvention(ValidatableYamlNode):
 
     @property
     def attributes(self):
+        if not hasattr(self, "attrs_by_name"):
+            return []
+
+        return list(
+            filter(
+                lambda attr: not AttributeType.is_template_type(attr.attr_type),
+                list(self.attrs_by_name.values()),
+            )
+        )
+
+    @property
+    def attribute_templates(self):
+        if not hasattr(self, "attrs_by_name"):
+            return []
+
+        return list(
+            filter(
+                lambda attr: AttributeType.is_template_type(attr.attr_type),
+                list(self.attrs_by_name.values()),
+            )
+        )
+
+    @property
+    def attributes_and_templates(self):
         if not hasattr(self, "attrs_by_name"):
             return []
 
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -16,7 +16,7 @@ import sys
 import typing
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Dict, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 from ruamel.yaml import YAML
 
@@ -112,34 +112,26 @@ class BaseSemanticConvention(ValidatableYamlNode):
 
     @property
     def attributes(self):
-        if not hasattr(self, "attrs_by_name"):
-            return []
-
-        return list(
-            filter(
-                lambda attr: not AttributeType.is_template_type(attr.attr_type),
-                list(self.attrs_by_name.values()),
-            )
-        )
+        return self._get_attributes(False)
 
     @property
     def attribute_templates(self):
-        if not hasattr(self, "attrs_by_name"):
-            return []
-
-        return list(
-            filter(
-                lambda attr: AttributeType.is_template_type(attr.attr_type),
-                list(self.attrs_by_name.values()),
-            )
-        )
+        return self._get_attributes(True)
 
     @property
     def attributes_and_templates(self):
+        return self._get_attributes(None)
+
+    def _get_attributes(self, templates: Optional[bool]):
         if not hasattr(self, "attrs_by_name"):
             return []
 
-        return list(self.attrs_by_name.values())
+        return [
+            attr
+            for attr in self.attrs_by_name.values()
+            if templates is None
+            or templates == AttributeType.is_template_type(attr.attr_type)
+        ]
 
     def __init__(self, group):
         super().__init__(group)
--- a/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
+++ b/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
@@ -102,10 +102,8 @@ class MarkdownRenderer:
         attr_type = (
             "enum"
             if isinstance(attribute.attr_type, EnumAttributeType)
-            else attribute.attr_type
+            else AttributeType.get_instantiated_type(attribute.attr_type)
         )
-        if AttributeType.is_template_type(attr_type):
-            attr_type = AttributeType.get_core_template_type(attr_type)
         description = ""
         if attribute.deprecated and self.options.enable_deprecated:
             if "deprecated" in attribute.deprecated.lower():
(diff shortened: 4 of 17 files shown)


jsuereth (approved):
This LGTM as a way forward

lmolkova:
Nice!!
I wonder if we can achieve the same with less new concepts:

```yaml
groups:
  - id: http.request.headers
    type: attribute_group
    prefix: http.request.header 
    brief: 'This group defines dynamic attributes in `http.request.headers` namespace'
    attributes:
      - id: key
        type:  parameter #or 'wildcard', or 'template', or 'dynamic'   <<---- THIS IS NEW TYPE
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```

or even this:

```yaml
groups:
  - id: http.request.headers
    type: wildcard_attribute_group <<---- THIS IS NEW TYPE
    prefix: http.request.header 
    brief: 'This group registers a namespace for HTTP request headers'
```

I personally prefer the latter because we don't need to define any attributes. I also think that ability to register a namespace without any specific attributes is useful beyond wildcard attributes.

AlexanderWert:
> I wonder if we can achieve the same with fewer new concepts

@lmolkova 
I agree with that! Maybe we can find a way to represent it with less new concepts in the yaml files. However, I think, in the implementation of the generator we still should separate those concepts. (i.e. template attributes should not be included when requesting the list of all attributes. Otherwise existing code generation logic could break.)

> I personally prefer the latter because we don't need to define any attributes (and specify levels, or type which make no sense in this case).

I think `requirement level` and `type` could still be reasonable for this kind of attributes. For example the [`http.request.header.<key>`](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#http-request-and-response-headers) has a defined `string` type and the requirement-level `opt-in`. 

Or the [`db.elasticsearch.path_parts.<key>`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md) has a `string` type and requirement level `conditionally required`. So, how about we try to model it as a special kind of attribute (similar to your first proposal)?


**Regarding the first proposal**:
I like the idea. There's only one issue with that. It implies that the `id` of that attribute is the "dynamic" part of the attribute name (in the above example `key`) that would be put into angle brackets. While it works in the above example, it might be problematic in other places. 
For example, if we would add `http.request.header.<key>` in [this yaml file](https://github.com/open-telemetry/semantic-conventions/blob/c6efa4b2b44f9950b332a46311834e23b85bf979/model/trace/http.yaml#L10), based on the above proposal the `id` would be `request.header.key`, but only `key` within that id is the dynamic part. 

So, what do you think about the following counter proposal?:

```
groups:
  - id: http.trace.http.common
    type: attribute_group
    prefix: http 
    brief: 'This group defines dynamic attributes in `http.request.headers` namespace'
    attributes:
      - id: request.header
        type:  string
        template_name: key # <<---- THIS IS NEW
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```

So, instead of a `boolean` template property, having a `template_name` property that specifies the name of the dynamic part. If `template_name` is specified it indicates it's a template attribute, if it's not specified (i.e. null) then it's a normal attribute. With that it would be only one additional field, while it still would be possible to address all of the points I described above.

WDYT?

AlexanderWert:
> there seems to be no logical reason to render them out-of-order

I agree for rendering in the markdown files: they should be treated very similar / same as normal attributes in the markdown files.

However, wouldn't we explicitly want to treat them separately for code generation because they will result in completely different constructs?

For example in Java, normal attributes result in [constants in this file](https://github.com/open-telemetry/opentelemetry-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java), however, template attributes have a different construct, which is a helper class, [like done here manually](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/b3932c7089dc07b815063ff4ae78db11fa8b3a8a/instrumentation-api-semconv/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/CapturedHttpHeadersUtil.java#L17).

Wouldn't that imply that we actually want the code rendering for template attributes to happen out-of-order (or at least from a different set as the normal attributes)?

Otherwise, we'd need to explicitly exclude template attributes from places [like in this template](https://github.com/open-telemetry/opentelemetry-java/blob/9ecc6f08d8de95e3229d55804b62b02f7877b5d8/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2#L65C24-L65C34).

Maybe I didn't fully get yet the benefit or cases for:

> easier to handle non-template and template attributes in the same loop

Oberon00:
For the case of Java it might indeed be easier, but I think many languages just produce string constants.

> Maybe I didn't fully get yet the benefit or cases for:
>
>>    easier to handle non-template and template attributes in the same loop

No, I think you got it. It's not that bit of a benefit (e.g. you can apparently use `+` in the template to combine lists, while on the other hand you can also use filters like reject/rejectattr/select/[selectattr](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.rejectattr) to separate them nicely), more a matter of taste what should be the default.

AlexanderWert:
> more a matter of taste what should be the default.

I'm fine with both. The main motivation to keep it separate was to not introduce braking changes with that. 

How about inverting your proposal? So, not separating the concepts, but instead of opting-out rather opt-in into rendering template-attributes with something like `--include-template-attributes` (instead of `--suppress-template-attributes`).

That would not separate those but also prevent breaking changes (or the need to change invocation parameters to keep the current behavior).

Oberon00:
I was thinking of that as well, but that would make the tool suppress information by default and you have to know about the flag, otherwise everything will silently work but be reduced. I'd prefer the separate lists over that I think, but it's a very fine like at this point

AlexanderWert:
@lmolkova , @trask 
Do you have opinions on [my proposal](https://github.com/open-telemetry/build-tools/pull/186#issuecomment-1628321598) and the discussion above?

Otherwise, I'd adapt this PR according to that proposal, WDYT?

lmolkova:
@AlexanderWert it looks great! 

One thing I'm slightly  concerned about is `id` field.

I wonder if it can match the full attribute name, i.e.

```yaml
groups:
  - id: http.trace.http.common
    type: attribute_group
    prefix: http 
    brief: 'This group defines dynamic attributes in `http.request.headers` namespace'
    attributes:
      - id: request.header.key #or request.header.{key}, request.header.[key], etc
        type:  string
        template_name: key # <<---- THIS IS NEW
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```
otherwise, we're defining a namespace `request.header` under `attributes` section and it does not feel right.

Oberon00:
How about we use something different from `id`? For example we could use `prefix` again or `namespace`. We might then not even need `template_name` (wouldn't it be set to "key" 99% of the time?)

Oberon00:
Agreed, it seems to be a good idea to mark the `id` already somehow, either with special syntax or a different YAML property name.

How about using `prefix` instead of `id`, not using any trailing `*` or similar and not using something like template_name? The name of the key will probably be "key" 99,9% of the time anyways

AlexanderWert:
It seems that the tooling heavily relies on the ID field and also that is not allowed to contain special characters like `*`.

So I see the following options:
1. Keep the ID field, but use some special syntax for template types, for example: `request.header._` instead of `request.header.*`
2. Not specifying the ID in the yaml file but deriving an artificial ID in the tooling, so it's transparent to the users. 

> ```yaml
> groups:
>     attributes:
>       - id_template:  request.header.{key} // or request.header.[key], etc # <<---- THIS IS NEW
>         type:  string
>         brief: >
>           HTTP request headers, `<key>` being the normalized HTTP Header name
>           (lowercase, with - characters replaced by _), the value being the header values.
>         examples: '`http.request.header.content_type=["application/json"]`'
> ```

I like this proposal. It's compact and at the same time contains all the information needed.

If there are no objections, I'm going to adapt the PR to implement that using alternative `2.` from above.

AlexanderWert:
The problem with omitting the `id` is that it's not possible to reference these attributes from other places / yaml files anymore.

So how about @lmolkova 's second proposal but keeping the `id` instead of `id_template`. That basically means that the format of the `id` implicitly indicates whether it's a template type or a normal attribute:

```
groups:
    attributes:
      - id:  request.header.<key> //  <<---- SPECIAL FORMAT that allows the tooling to differentiate
        type:  string
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```

Oberon00:
> The problem with omitting the id is that it's not possible to reference these attributes from other places 

It would be possible to adapt this so that for referencing (and also unique-ness checks!) it behaves just like "id". Might be too cumbersome / complicate the code too much to be worth it though.

AlexanderWert:
@lmolkova , @Oberon00 

I adapted the PR to implement this approach:

```yaml
groups:
    attributes:
      - id:  request.header.<key> //  <<---- SPECIAL FORMAT that allows the tooling to differentiate
        type:  string
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```


I think it's the least invasive still the `id` is unique and indicates that it's a special type of attribute.

- Internally the list of attributes excludes the template attributes to not break existing code generation
- Table generation works for the actual template attribute definitions as well as for references to template attributes

Please have another look and if you're fine with that let's get this merged so we can move those parts to YAML as well.

Thanks!

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -84,7 +87,7 @@ def is_enum(self):
 
     @staticmethod
     def parse(
-        prefix, semconv_stability, yaml_attributes
+        prefix, semconv_stability, yaml_attributes, parse_template_attributes=False
```
Oberon00, line 90: I think we should rather not mix filtering and parsing. If we want to filter them out, we can do so after parsing

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -116,7 +120,7 @@ def parse(
                 msg = "At least one of id or ref is required."
                 raise ValidationError.from_yaml_pos(position, msg)
             if attr_id is not None:
-                validate_id(attr_id, position_data["id"])
+                validate_attribute_id(attr_id, position_data["id"])
                 attr_type, brief, examples = SemanticAttribute.parse_id(attribute)
                 if prefix:
                     fqn = f"{prefix}.{attr_id}"
```
Oberon00, line 126: The key name now is part of the fqn. This is IMHO wrong. The fqn, for the purpose of uniqueness-checks and referencing should be independent of the key name. E.g. it should be invalid to have both `request.header.<key>` and `request.header.<name>`, or even both `request.header` and `request.header.<key>`.

I think it will be cleaner to implement if the id is just the ID of the namespace and the fact that it's a template is indicated with some other way.

If you want to stay with the indicator in the ID, consider using `*` instead of `<customkeyname>`, and adding special handling to catch the case when both `name` and `name.*` exist.

AlexanderWert: > I think it will be cleaner to implement if the id is just the ID of the namespace and the fact that it's a template is indicated with some other way.

But isn't that exactly what I proposed before and @lmolkova was concerned about:

> otherwise, we're defining a namespace request.header under attributes section and it does not feel right

I'm really fine with both. Even just keeping the `id` as `request.header`. The implication of that would be that `request.header` is identifying this set of attributes, but it cannot be used as a namespace for other attributes, which is fine IMHO.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
AlexanderWert: agree

AlexanderWert:
> otherwise, we're defining a namespace `request.header` under `attributes` section and it does not feel right.

@lmolkova Actually, we are not defining a namespace with that, but a `template attribute` with the unique id `http.request.header`. According to the attributes naming rules, this prohibits the usage of `http.request.header` as a namespace. Or am I missing something?

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
AlexanderWert: The more I think about it, the more I like your idea, @Oberon00 

> Another idea would be to use `type: namespace[string]` or similar, but that would be difficult to implement I think.

Oberon00: > But isn't that exactly https://github.com/open-telemetry/build-tools/pull/186#issuecomment-1628321598 and @lmolkova was concerned about?:

I think it is. @lmolkova is this a big concern or just a nice-to-(not-)have?

AlexanderWert:
Based on the feedback, the newest update in this PR contains the following variant:

```yaml
groups:
    attributes:
      - id:  request.header
        type:  template[string[]]
        brief: >
          HTTP request headers, `<key>` being the normalized HTTP Header name
          (lowercase, with - characters replaced by _), the value being the header values.
        examples: '`http.request.header.content_type=["application/json"]`'
```

jsuereth (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
lmolkova: Agree, we can validate namespace vs attribute name reuse with tooling given we have a special template type, no concern from my side.

lmolkova (commented):
Looks great!

Could you please also update a few things?
- changelog
- docs: https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/syntax.md#type
- schema definition: https://github.com/open-telemetry/build-tools/blob/main/semantic-conventions/semconv.schema.json#L282

I'm also curious how code generation will work on this - I'll try later today and share.

AlexanderWert:
@lmolkova 

I added the requested changes:

- changelog
- docs
- schema definition

In addition I also added the capability to render code (in a way that it should not affect existing code generation templates). Also added a test for the code generation of template attributes.

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -74,6 +76,14 @@ def import_attribute(self):
     def inherit_attribute(self):
         return replace(self, inherited=True)
 
+    @property
+    def core_type(self):
```
Oberon00, line 80: If we are using template terminology then maybe `get_instantiated_type`? With "core", I would expect to get `int` for `int[]`, and what about enums?

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -263,19 +270,32 @@ def parse_id(attribute):
...
+            not in (
+                "boolean",
+                "boolean[]",
+                "int",
+                "int[]",
+                "double",
+                "double[]",
+            )
         ):
             if not examples:
```
Oberon00, line 285: Examples are a good point. You'd want to show the key in the example for templates as well. But that can be done i a follow-up / is a nice-to-have

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -358,6 +378,22 @@ def is_simple_type(attr_type: str):
...
+        if matchObj is not None:
+            return AttributeType.is_simple_type(matchObj.group(1))
+        return False
+
+    @staticmethod
+    def get_core_template_type(attr_type: str):
+        matchObj = TEMPLATE_TYPE_RE.fullmatch(attr_type)
+        if matchObj is not None:
+            return matchObj.group(1)
+        raise ValidationError(0, 0, f"Cannot retrieve core type from type: {attr_type}")
```
Oberon00, line 395: I don't remember anymore: I the 0, 0 caught and reraised with the proper line number later? Or should we rather pass on the location to fill it out here?

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -114,6 +115,30 @@ def attributes(self):
         if not hasattr(self, "attrs_by_name"):
             return []
 
+        return list(
+            filter(
+                lambda attr: not AttributeType.is_template_type(attr.attr_type),
+                list(self.attrs_by_name.values()),
+            )
+        )
```
Oberon00, line 123:
```suggestion
        return [attr for attr in self.attrs_by_name.values() if not AttributeType.is_template_type(attr.attr_type)]
```

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -114,6 +115,30 @@ def attributes(self):
...
+    def attribute_templates(self):
+        if not hasattr(self, "attrs_by_name"):
+            return []
+
+        return list(
+            filter(
+                lambda attr: AttributeType.is_template_type(attr.attr_type),
+                list(self.attrs_by_name.values()),
+            )
+        )
```
Oberon00, line 135: Extract a helper method: `def _get_attributes(self, *, templates: bool)` that is called from `attribute_templates` and `attributes` since they only differ by checking whether is_template_type is True vs False.

semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
```diff
@@ -325,19 +330,35 @@ def to_markdown_enum(self, output: io.StringIO):
...
         """
         Method to render in markdown an attribute id. If the id points to an attribute in another rendered table, a
         markdown link is introduced.
         """
+        diff = self.get_attr_reference_diff(attribute_id)
+        if diff is not None:
+            return f"[`{attribute_id}`]({diff})"
+        return f"`{attribute_id}`"
+
+    def get_attr_reference_diff(self, attribute_id):
```
Oberon00, line 353: You didn't introduce the term "diff" here but I think it's confusing. Probably "relpath" should be used instead of "diff"?

semantic-conventions/syntax.md
```diff
@@ -73,15 +73,20 @@ attributes ::= (id type brief examples | ref [brief] [examples]) [tag] [stabilit
...
      |   "int"
      |   "double"
      |   "boolean"
      |   "string[]"
      |   "int[]"
      |   "double[]"
      |   "boolean[]"
-     |   enum
+
+template_type ::= "template[" simple_type "]"
```
Oberon00, line 89:
```suggestion
template_type ::= "template[" simple_type "]"  # As a single string
```

semantic-conventions/syntax.md
```diff
@@ -213,17 +218,18 @@ Attribute groups don't have any specific fields and follow the general `semconv`
...
+  * _primitive and array types as string literals:_
+    * `"string"`: String attributes.
+    * `"int"`: Integer attributes.
+    * `"double"`: Double attributes.
+    * `"boolean"`: Boolean attributes.
+    * `"string[]"`: Array of strings attributes.
+    * `"int[]"`: Array of integer attributes.
+    * `"double[]"`: Array of double attributes.
+    * `"boolean[]"`: Array of booleans attributes.
+  * _template type as string literal:_ `"template[<PRIMITIVE_OR_ARRAY_TYPE>]"` (See later)
```
Oberon00, line 232:
```suggestion
  * _template type as string literal:_ `"template[<PRIMITIVE_OR_ARRAY_TYPE>]"` (See [below](#template-type))
```

semantic-conventions/syntax.md
```diff
@@ -333,7 +339,38 @@ fields are present in the current attribute definition, they override the inheri
 #### Type
 
 An attribute type can either be a string, int, double, boolean, array of strings, array of int, array of double,
-array of booleans, or an enumeration. If it is an enumeration, additional fields are required:
+array of booleans, a template type or an enumeration. 
+
+##### Template type
+
+A template type attribute represents a _key-value set_ of attributes with a common attribute name prefix. The syntax for defining template type attributes is the following:
```
Oberon00, line 346:
```suggestion
A template type attribute represents a _map/dictionary_ of attributes with a common attribute name prefix. The syntax for defining template type attributes is the following:
```

key-value set is an unusual term, and I think it would imply that you can have the same key multiple times, as long as the values are different.

semantic-conventions/syntax.md
```diff
@@ -333,7 +339,38 @@ fields are present in the current attribute definition, they override the inheri
...
+
+`type: template[<PRIMITIVE_OR_ARRAY_TYPE>]`
+
+The `<PRIMITIVE_OR_ARRAY_TYPE>` is one of the above-mentioned primitive or array types (_not_ an enumeration) and specifies the type of the `value` in the key-value pairs.
```
Oberon00, line 350:
```suggestion
The `<PRIMITIVE_OR_ARRAY_TYPE>` is one of the above-mentioned primitive or array types (_not_ an enum) and specifies the type of the `value` in the key-value pairs. 
```

semantic-conventions/syntax.md
```diff
@@ -333,7 +339,38 @@ fields are present in the current attribute definition, they override the inheri
...
+```yaml
+groups:
+  - id: trace.http.common
+    type: attribute_group
+    brief: "..."
+    attributes:
+      - id: http.request.header
+        type: template[string[]]
+        brief: >
+          HTTP request headers, `<key>` being the normalized HTTP Header name (lowercase, with `-` characters replaced by `_`), the value being the header values.
```
Oberon00, line 363:
```suggestion
          HTTP request headers, the key being the normalized HTTP header name (lowercase, with `-` characters replaced by `_`), the value being the header values.
```

Oberon00 (approved):

semantic-conventions/semconv.schema.json
```diff
@@ -287,7 +287,15 @@
 								"string[]",
 								"int[]",
 								"double[]",
-								"boolean[]"
+								"boolean[]",
+								"template[string]",
```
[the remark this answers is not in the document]
AlexanderWert: Kept it as is for now to avoid complexity. Agree with:

> But it's a technical debt that may need to be paid someday

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
```diff
@@ -27,6 +27,8 @@
     validate_values,
 )
 
+TEMPLATE_TYPE_RE = re.compile("template\\[([a-z\\[\\]]+)\\]")
```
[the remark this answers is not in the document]
AlexanderWert: Replaced with startswith()

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
AlexanderWert: done

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
AlexanderWert: > But that can be done i a follow-up / is a nice-to-have

Agree

semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py
AlexanderWert: Actually, we don't need it here, because it will be checked during parsing anyways.

semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
AlexanderWert: renamed it

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
AlexanderWert: done

lmolkova (approved):

trask (approved):

AlexanderWert:
@lmolkova can this be merged? Or anything missing?

AlexanderWert:
@open-telemetry/specs-semconv-maintainers Can we merge this?

joaopgrassi (approved):

PR #189, opened 2023-07-24
Change default stability level to experimental

### LINKED ISSUE #188: Fix default stability level
If unspecified, the current implicitly assumed value for `stability` is `STABLE`.

As discussed in the semantic conventions WG group (meeting on 2023-07-24), this is not a sane default and should be changed to `EXPERIMENTAL` instead.

cc @open-telemetry/specs-semconv-approvers

arminru:
Fixes #188

### DIFF
--- a/semantic-conventions/CHANGELOG.md
+++ b/semantic-conventions/CHANGELOG.md
@@ -4,6 +4,11 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
+## v0.20.0
+
+- Change default stability level to experimental
+  ([#189](https://github.com/open-telemetry/build-tools/pull/189))
+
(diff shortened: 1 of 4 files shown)


Oberon00 (approved):

lmolkova (approved):
Thank you!

arminru:
@open-telemetry/specs-semconv-maintainers @open-telemetry/specs-semconv-approvers Please review 🙂 
Some might consider this change breaking but it shouldn't be surprising to anyone nor do I believe this is already being relied on while wrongly assuming everything is stable, so I think this change is still fine at this point.

jsuereth (approved):

trask (approved):

PR #190, opened 2023-07-26
Added `omit_requirement_level` option for markdown table rendering

AlexanderWert:
When rendering attributes tables outside specific semantic conventions (e.g. in an attributes dictionary), we need a way to omit rendering the `requirement_level` column in the table.

See this proposal:
- https://github.com/open-telemetry/semantic-conventions/issues/197

This PR introduces an optional `omit_requirement_level` parameter in the markdown `semconv_selector` syntax that indicates the table rendering logic to omit rendering the `requirement_level` column in the corresponding attributes table.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
+++ b/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
@@ -45,6 +45,7 @@ class RenderContext:
         self.is_full = False
         self.is_remove_constraint = False
         self.is_metric_table = False
+        self.is_omit_requirement_level = False
         self.group_key = ""
         self.enums = []
         self.notes = []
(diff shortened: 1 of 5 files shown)


Oberon00:
I dont't think this should be a markdown-only option. Code generators would still see the requirement levels. IMHO, this needs to be somehow modelled in the YAML.

AlexanderWert:
> IMHO, this needs to be somehow modelled in the YAML.

I agree with that! But it sounds like a huge refactoring, so I was wondering if we could do this pragmatic step as a intermediate step to get https://github.com/open-telemetry/semantic-conventions/issues/197 started before changing the yaml syntax significantly (so, as part of step 1 in https://github.com/open-telemetry/semantic-conventions/issues/197#issuecomment-1649800369).

> I dont't think this should be a markdown-only option. Code generators would still see the requirement levels.

Just curious: are code generators using the requirement level at all? IIUC, they are generating the language-specific constants for the attributes (which do not contain the requirement level in any form). Am I missing something?

AlexanderWert:
> IMHO, this needs to be somehow modelled in the YAML.
@Oberon00 See this initial proposal: https://github.com/open-telemetry/build-tools/pull/191

Oberon00 (approved):
Approved because the code looks fine and I don't want to block this. I have no opinion on whether we want to take this intermediate step or should try for a proper solution from the start.

joaopgrassi:
I created #202 so we don't lose track to properly implement the final solution in YAML.

AlexanderWert:
> Can we have some tests just to make sure things are working as expected? We should have examples there that can make it easier (I think).

Will add some

AlexanderWert:
@joaopgrassi I added a test for the table rendering while omitting the requirement-level column.

I think we are good to merge this and then do the release of the build-tools?

joaopgrassi (approved):

arminru (approved):

PR #201, opened 2023-09-05
Fixed conditionally_required requirement_level.

jerbly:
The `conditionally_required` option for `requirement_level` has an error in the schema.json. This causes the vscode extension to flag an error when there is none. Here's an example from http-common: 
image

The fix is simply to change the properties to match `conditionally_required`.


joaopgrassi (approved):

AlexanderWert (approved):

Oberon00 (approved):

lmolkova (approved):

jerbly:
Hey. Thanks for all the approvals! What's the next step to get this merged?

arminru (approved):

PR #204, opened 2023-09-17
Inherit overridden attribute properties

lmolkova:
In some cases, we want to have the following schema:
- `server.address` is a general-purpose attribute defined in `server` conv

```yaml
  - id: server
    prefix: server
    attributes:
      - id: address
        brief: 'Domain name.'
       ....
```
- `http` semconv wants to define it more narrowly. E.g. update a brief or add HTTP-specific note
```yaml
- ref: server.address
  brief: Server component of Host header.
```
- Then HTTP version of this attribute can be used in HTTP traces and metrics.

Today it works only if semconv extends base HTTP one without mentioning the attribute like
```yaml
  - id: http.client.request.duration.metric
    type: metric
    metric_name: http.client.request.duration
   ...
    extends: http.client.attributes
```

Then an attribute is populated from parent semconv as is.

However, if child semconv specifies an attribute even further, it fills it in from the original definition, without considering parent hierarchy first.

I.e. 
```yaml
  - id: http.client.spans
    type: span
   ...
    extends: http.client.attributes
    attributes:
      - ref: server.address
        sampling_relevant: true
```

resolves to 

|---|---|---|---|---|
| [`server.address`](input_server.md) | string | Domain name. | `foo` | Recommended |`

The change implemented here walks up the `extended` hierarchy and fills in properties in the attribute first.

### DIFF
--- a/semantic-conventions/src/tests/data/markdown/ref_extends/http.yaml
+++ b/semantic-conventions/src/tests/data/markdown/ref_extends/http.yaml
@@ -1,0 +1,24 @@
+groups:
+  - id: http.client.attributes
+    type: attribute_group
+    brief: 'This document defines semantic conventions for HTTP client attributes.'
+    attributes:
+      - ref: server.address
+        requirement_level: required
+        brief: Server component of Host header.
+
+  - id: http.client.spans
+    type: span
+    brief: 'This document defines semantic conventions for HTTP client Spans.'
+    extends: http.client.attributes
+    attributes:
+      - ref: server.address
+        sampling_relevant: true
+
+  - id: http.client.request.duration.metric
+    type: metric
+    metric_name: http.client.request.duration
+    brief: "Measures request duration."
+    instrument: histogram
+    unit: "s"
+    extends: http.client.attributes
--- a/semantic-conventions/src/tests/data/markdown/ref_extends/server.yaml
+++ b/semantic-conventions/src/tests/data/markdown/ref_extends/server.yaml
@@ -1,0 +1,10 @@
+groups:
+  - id: server
+    type: attribute_group
+    prefix: server
+    brief: 'This document defines semantic conventions for common server attributes.'
+    attributes:
+      - id: address
+        type: string
+        brief: 'Domain name.'
+        examples: 'foo'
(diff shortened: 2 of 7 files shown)


jsuereth (approved):

AlexanderWert (approved):

Oberon00 (approved):
So basically, instead of starting from the primary definition, we start from the most-derived defintion, if we inherit it already? Sounds good 👍 

Could you add this to the CHANGELOG, please?

semantic-conventions/src/tests/data/markdown/ref_extends/server.yaml
```diff
@@ -0,0 +1,10 @@
+groups:
+  - id: server
+    type: attribute_group
+    prefix: server
+    brief: 'This document defines semantic conventions for common server attributes.'
+    attributes:
+      - id: address
+        type: string
+        brief: 'Domain name.'
```
joaopgrassi, line 9: Would be nice if the "briefs" contained where in the hierarchy they are, so it's easier to understand the expect/actual results. like `(Grandparent) domain name` or smt similar.

semantic-conventions/src/tests/data/markdown/ref_extends/http.yaml
```diff
@@ -0,0 +1,24 @@
+groups:
+  - id: http.client.attributes
+    type: attribute_group
+    brief: 'This document defines semantic conventions for HTTP client attributes.'
+    attributes:
+      - ref: server.address
+        requirement_level: required
+        brief: Server component of Host header.
```
joaopgrassi, line 8: Should we also add a `note` here to make sure that is also working as expected? Also the requirement level message I guess.

lmolkova:
> So basically, instead of starting from the primary definition, we start from the most-derived defintion, if we inherit it already? Sounds good 👍

yep!

> Could you add this to the CHANGELOG, please?

done

joaopgrassi (approved):

semantic-conventions/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
+- Inherit referenced attribute properties from `extend`
+  hierarchy before falling back to original attribute definition.
```
Oberon00, line 8: "from hierarchy" does not tell much to me. Maybe:

```suggestion
- When an attribute is referenced using `ref:` from a group that already inherits the attribute with `extends:`, resolve the reference to the closest inherited attribute instead of the primary definition.
```

arminru (approved):

PR #205, opened 2023-09-18
Sort attributes by name, not by source

lmolkova:
How attributes are added to semconv is an implementation detail, rendered markdown should not depend on it.

This change sorts attributes in final md table alphabetically. 

This would also reduce diffs in semconv and help us review uninteresting changes when hierarchy refactorings cause massive changes because of attributes moving in the table.

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -437,7 +437,9 @@ class SemanticConventionSet:
             # By induction, parent semconv is already correctly sorted
             # but the combination of parent and current attributes is not
             if parent_attributes or semconv.attributes:
-                semconv.attrs_by_name = SemanticConventionSet._sort_attributes_dict(parent_attributes)
+                semconv.attrs_by_name = SemanticConventionSet._sort_attributes_dict(
+                    parent_attributes
+                )
 
         elif semconv.attributes:  # No parent, sort of current attributes
             semconv.attrs_by_name = SemanticConventionSet._sort_attributes_dict(
@@ -455,9 +457,7 @@ class SemanticConventionSet:
         :param attributes: Dictionary of attributes to sort
         :return: A sorted dictionary of attributes
         """
-        return dict(
-            sorted(attributes.items())
-        )
+        return dict(sorted(attributes.items()))
 
     def _populate_anyof_attributes(self):
         any_of: AnyOf
--- a/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
+++ b/semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
@@ -126,12 +126,15 @@ class BaseSemanticConvention(ValidatableYamlNode):
         if not hasattr(self, "attrs_by_name"):
             return []
 
-        return sorted([
-            attr
-            for attr in self.attrs_by_name.values()
-            if templates is None
-            or templates == AttributeType.is_template_type(attr.attr_type)
-        ], key=lambda attr: attr.fqn)
+        return sorted(
+            [
+                attr
+                for attr in self.attrs_by_name.values()
+                if templates is None
+                or templates == AttributeType.is_template_type(attr.attr_type)
+            ],
+            key=lambda attr: attr.fqn,
+        )
 
     def __init__(self, group):
         super().__init__(group)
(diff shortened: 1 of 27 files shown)


lmolkova:
> I would prefer either using a sensible order in the YAML (alphabetical might or might not be the best, there might be closely related attributes that are best put together or where there is another logical order), or one could argue that this should instead be a lint (error if input is not correctly sorted)

why though? what's the benefit to someone reading the semconv? 

For authors, there is always a possibility of rendering individual tables per group of attributes. Having a predictable order also helps to review things.

As a user, I want to be able to skim through the table a find my attribute in certain place, not knowing where author of semconv wanted to put it.

Oberon00:
> Having a predictable order also helps to review things.

Definitely. I was under the impression that currently the order is the same as in the YAML? Do we have some random order?

lmolkova:
> Definitely. I was under the impression that currently the order is the same as in the YAML? Do we have some random order?

the order is random since:
1. afaik nobody puts any meaning into it.
2. even if we did, `extends` are implementation detail, but once we're adding one, we no longer control the relative order for defined and extended attributes.

either way, the order author had in mind is not obvious to readers, who end up with de-facto randomly ordered attributes 
PTAL here (or anywhere else)

https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#common-attributes

lmolkova:
One more observation: since we have a lot of namespaces now, most attributes are going to be referenced from somewhere and namespaces provide all the reasonable grouping.

Later on, we can consider adding sorting by requirement level and whatnot, but current random thing hurts a lot during review.

AlexanderWert (approved):

joaopgrassi (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -453,9 +456,7 @@ def _sort_attributes_dict(
         :param attributes: Dictionary of attributes to sort
```
Oberon00, line 456: The function docs (cut off by GH context) claim "First  imported, and then defined attributes.". I think this is not true anymore, now it's simply alphabetical? Also maybe check if some other follow-up simplifications/comment-corrections are possible now.

lmolkova: good point! I simplified code a bit

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -431,32 +431,15 @@ def _populate_extends_single(self, semconv, unprocessed):
...
-                semconv.attrs_by_name = parent_attributes
-        elif semconv.attributes:  # No parent, sort of current attributes
-            semconv.attrs_by_name = SemanticConventionSet._sort_attributes_dict(
-                semconv.attrs_by_name
-            )
+
+            parent_attributes.update(semconv.attrs_by_name)
+            semconv.attrs_by_name = parent_attributes
+
+        semconv.attrs_by_name = dict(sorted((semconv.attrs_by_name.items())))
```
Oberon00, line 438:
```suggestion
        semconv.attrs_by_name = dict(sorted(semconv.attrs_by_name.items()))
```

Oberon00 (approved):

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -431,32 +431,15 @@ def _populate_extends_single(self, semconv, unprocessed):
...
+
```
AlexanderWert, line 439: I just tested this. It seems that the sorting is not being applied on attribute references.

In particular, the test in `tests/data/markdown/ref` still passes successfully with the following expected output (which is not being sorted alphabetically):

```
|---|---|---|---|---|
| `rpc.service` | string | The service name, must be equal to the $service part in the span name. | `EchoService` | Required |
| [`net.peer.name`](input_general.md) | string | override brief. [1] | `example.com` | Opt-In |
...
```
Here, `net.*` attributes are listed _after_ the `rpc.service` attribute just because those are references.

- I think, we need to adapt that test with the expected sorted output and **fix the sorting to work for references** as well
- An explicit test (or adapting the existent test) for making sure the sorting also works for template-type attributes would be good as well

AlexanderWert (changes requested):

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
lmolkova: Thanks a lot @AlexanderWert! I believe I fixed it.

AlexanderWert (approved):

arminru (approved):

PR #206, opened 2023-09-27
Fix referencing template attributes

AlexanderWert:
In https://github.com/open-telemetry/build-tools/pull/186 there was an oversight that template-type attributes cannot be referenced in other semantic conventions.

This PR fixes this bug and extends the corresponding test with a referenced template attribute.


AlexanderWert:
@open-telemetry/specs-semconv-approvers , @open-telemetry/specs-semconv-maintainers 

PTAL and let's get this fix in. https://github.com/open-telemetry/semantic-conventions/pull/208 depends on this fix.

lmolkova (approved):
LGTM, but wanted to make sure we also support including template attributes by extending another semconv like 

```yaml
  - id: foo
    attributes:
      - id: template
        type: template[string]

  - id: bar
    extends:  foo
        type: template[string]
```

AlexanderWert:
> LGTM, but wanted to make sure we also support including template attributes by extending another semconv like

@lmolkova Good point! I fixed that now as well and extended the test to cover it.

joaopgrassi (approved):

PR #213, opened 2023-10-12
Adds an option to render briefs and notes of semconv attribute groups in md

### LINKED ISSUE #15: "brief"s and "note"s of groups should be (optionally?) included in Markdown output

AlexanderWert:
Fixes open-telemetry/semantic-conventions#1086

### DIFF
--- a/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
+++ b/semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
@@ -46,6 +46,8 @@ class RenderContext:
         self.is_remove_constraint = False
         self.is_metric_table = False
         self.is_omit_requirement_level = False
+        self.is_render_group_brief = False
+        self.is_render_group_note = False
         self.group_key = ""
         self.enums = []
         self.notes = []
@@ -77,6 +79,8 @@ class MarkdownRenderer:
         "remove_constraints",
         "metric_table",
         "omit_requirement_level",
+        "render_group_brief",
+        "render_group_note",
     ]
 
     prelude = "<!-- semconv {} -->\n"
@@ -206,6 +210,12 @@ class MarkdownRenderer:
     def to_markdown_attribute_table(
         self, semconv: BaseSemanticConvention, output: io.StringIO
     ):
+        if self.render_ctx.is_render_group_brief:
+            output.write(semconv.brief + "\n\n")
+
+        if self.render_ctx.is_render_group_note:
+            output.write(semconv.note + "\n\n")
+
         attr_to_print = []
         for attr in semconv.attributes_and_templates:
             if self.render_ctx.group_key is not None:
@@ -517,6 +527,12 @@ class MarkdownRenderer:
         self.render_ctx.is_metric_table = "metric_table" in parameters
         self.render_ctx.is_omit_requirement_level = (
             "omit_requirement_level" in parameters
+        )
+        self.render_ctx.is_render_group_brief = (
+            "render_group_brief" in parameters
+        )
+        self.render_ctx.is_render_group_note = (
+            "render_group_note" in parameters
         )
 
         if self.render_ctx.is_metric_table:
--- a/semantic-conventions/src/tests/data/markdown/attribute_group/attribute_group.yaml
+++ b/semantic-conventions/src/tests/data/markdown/attribute_group/attribute_group.yaml
@@ -16,7 +16,7 @@ groups:
     extends: attributes
     prefix: "foo"
     brief: Derived attribute group
-    note: This is a groupt note.
+    note: This is a group note.
     attributes:
       - id: qux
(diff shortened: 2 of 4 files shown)


semantic-conventions/src/tests/data/markdown/attribute_group/attribute_group.yaml
```diff
@@ -16,6 +16,7 @@ groups:
     extends: attributes
     prefix: "foo"
     brief: Derived attribute group
+    note: This is a groupt note.
```
lmolkova, line 19: nit
```suggestion
    note: This is a group note.
```

semantic-conventions/src/tests/data/markdown/attribute_group/expected.md
```diff
@@ -6,3 +6,19 @@
...
+<!-- semconv attributes(render_group_brief) -->
+Attribute group brief.
+
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `foo.bar` | string | Attribute 1 | `baz` | Recommended: if available |
+<!-- endsemconv -->
+
+<!-- semconv derived_attributes(render_group_note) -->
+This is a groupt note.
```
lmolkova, line 19:
```suggestion
This is a group note.
```

semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
```diff
@@ -206,6 +210,12 @@ def write_table_header(self, output: io.StringIO):
     def to_markdown_attribute_table(
         self, semconv: BaseSemanticConvention, output: io.StringIO
     ):
+        if self.render_ctx.is_render_group_brief:
+            output.write(semconv.brief + "\n\n")
```
lmolkova, line 214: Is there a case when we don't want to render brief and note? I'd prefer it to be a default (and maybe even not configurable).

Also, I'm not sure if we support multiple configurations (like `full`, `tags` and `render_group_brief`). If not, we'd have a harder time using this in semconv repo.

lmolkova (commented):
Great idea! I would just prefer it to be the default non-configurable behavior.

semantic-conventions/src/opentelemetry/semconv/templating/markdown/__init__.py
AlexanderWert: > Is there a case when we don't want to render brief and note? I'd prefer it to be a default (and maybe even not configurable).

I can imagine cases where we might want to only render the table without brief and note. So, I think having it being configurable is a good thing. I'm up for making it enabled by default, though! Though, that would mean that all the existing markdown files would change once we update to a new version of the build-tools that would include that feature. 

> Also, I'm not sure if we support multiple configurations (like full, tags and render_group_brief). If not, we'd have a harder time using this in semconv repo.

Yes, IIUC that's possible. You can specify multiple configurations at once ([here's an example](https://raw.githubusercontent.com/open-telemetry/build-tools/main/semantic-conventions/src/tests/data/markdown/parameter_remove_constraint/input.md)). Some of the config options are logically mutually exclusive / contradicting , though, when used at once (like `full` and `remove_constraints`).

Maybe we should make the rendering of briefs and notes enabled when the `full` option is being used, and otherwise an opt-in? @lmolkova WDYT?

AlexanderWert: @lmolkova Any thoughts on the above?

Oberon00: IMHO:
* It should not be separately controllable. Just have one single option for both brief+note.
* It should be on by default eventually
* I wonder if a command line option wouldn't be better here, as I see it, it is mostly only for backwards compatibility that we have this option.

lmolkova: > @lmolkova Any thoughts on the above?

I see your point @AlexanderWert, thanks for the explanation. I don't really see a strong reason to go through all the steps to start populating briefs/notes and would rather:
- populate them by default right away. then send a big diff PR with just this change. If some notes/briefs are too bad, either fix them right away or disable them.
- maybe keep an option to not populate them on specific tables? 

WDYT?

Oberon00: The suggestions by @lmolkova is fine for me as well, that's the "more aggressive" option. So, always-on, no new options at all (neither command line nor per-table). Probably the most clean way.

jsuereth:
Closing this as work has moved to weaver project

Issue #212, opened 2023-10-08
[security] audit repository tooling


sakshi-1505:
The security SIG is looking to ensure that security tooling is setup consistently across the organization. As a result, we're asking maintainers to ensure the following tools are enabled in each repository:

- [ ] CodeQL enabled via [GitHub Actions](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/codeql-analysis.yml)
- [ ] Static code analysis tool (the collector uses govulncheck [https://pkg.go.dev/golang.org/x/vuln] on every build)
- [x] Repository security settings
  - [x] Security Policy ✅
  - [x] Security advisories ✅
  - [ ] Private vulnerability reporting ✅
  - [ ] Dependabot alerts ✅
  - [ ] Code scanning alerts ✅

Parent issue: https://github.com/open-telemetry/sig-security/issues/12

sakshi-1505:
@arminru Please confirm if vulnerability reporting, dependabot alerts are configured for the repo as I don't have enough access to see those. I do see we have majority of the repo in Python with a lil of Go but I don't see us using any staticcode checker for Python, do you folks have any suggestion in mind for the same? I know we can use [Mypy](https://github.com/python/mypy) or [Pylint](https://github.com/pylint-dev/pylint).

Meanwhile I will open a PR for CodeQL scanning with schedule same as other repos.

Oberon00:
@open-telemetry/technical-committee I think you are still the only ones that have access to view the mentioned settings. Please check.

@sakshi-1505 For Python we already use both mypy and pylint, and addtionally flake8 (and black + isort but these are probably less security relevant). https://github.com/open-telemetry/build-tools/blob/main/.github/workflows/semconvgen.yml Most of these could use an upgrade (we have dependabot PRs but with new linter errors that would need to be fixed), and there is also https://github.com/astral-sh/ruff which seems to be the new cool tool that would replace all 3 of pylint, flake8 and isort while being faster. So if somebody has time, there is always potential for improvement, but I think we have the basics covered (and note that the semantic convention generator is development tooling and not distributed to or used by end users)

sakshi-1505:
@Oberon00 please check #214

arminru:
@Oberon00 I looked into the repo settings mentioned in the issue description, adjusted them where needed, and ticked the boxes.

sakshi-1505:
Thanks a lot @arminru , can you please also review #214?

[...]

PR #214, opened 2023-10-15
feat: add codeql scheduled cronjob

sakshi-1505:
refers https://github.com/open-telemetry/build-tools/issues/212

### DIFF
--- a/.github/workflows/codeql_analysis.yml
+++ b/.github/workflows/codeql_analysis.yml
@@ -1,0 +1,36 @@
+on:
+  workflow_dispatch:
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │  ┌───────────── hour (0 - 23)
+    #        │  │ ┌───────────── day of the month (1 - 31)
+    #        │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        *  * * * *
+    - cron: '30 1 * * *'
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  CodeQL-Build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


sakshi-1505:
@codeboten Please check this

Oberon00 (dismissed):

.github/workflows/codeql_analysis.yml
```diff
@@ -0,0 +1,36 @@
...
+    #        ┌───────────── minute (0 - 59)
+    #        │  ┌───────────── hour (0 - 23)
+    #        │  │ ┌───────────── day of the month (1 - 31)
+    #        │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        │  │ │ │ │
+    #        *  * * * *
+    - cron: '30 1 * * *'
```
jpkrohling, line 13: I'm not sure we should have it periodically. Upon merge and on PRs should be sufficient.

Oberon00: I think a cron schedule is actually very useful. Often you will not introduce a security issue on merge, but it will pop up without code changes, simply because it is later discovered that it was already there but not known yet.

This is especially of interest if we don't merge anything for a while.

sakshi-1505: What should be the next step here folks?

sakshi-1505: @jpkrohling @Oberon00

jpkrohling: @Oberon00 has a good point, feel free to keep the scheduled runs.

jsuereth (commented):
In practice, this project is actually split between *three* different projects (at least):

Semconv generator, Protobuf and CPP tooling.

I'm concerned about two things with this repo:

1. Lack of attention from those named to own the repo (myself included)
2. Very broad set of responsibilities across OTEL with true owners (e.g. CPP tooling) being a different SiG.


For now, would you be willing to do codeql analysis separately on each subdirectory so we get a different alert/error notification and direct to the "true" owners?

.github/workflows/codeql_analysis.yml
jsuereth: @sakshi-1505 / @jpkrohling is this still relevant?

jsuereth:
I'm going to close this as obsolete - feel free to reopen/restart work if you'd like to continue!

PR #230, opened 2023-10-31
Update sampling-relevant MUST -> SHOULD

lmolkova:
Fix https://github.com/open-telemetry/semantic-conventions/issues/422

### DIFF
--- a/semantic-conventions/CHANGELOG.md
+++ b/semantic-conventions/CHANGELOG.md
@@ -4,6 +4,9 @@ Please update the changelog as part of any significant pull request.
 
 ## Unreleased
 
+- Rephrase and relax sampling-relevant description
+  ([#230](https://github.com/open-telemetry/build-tools/pull/230))
+
 ## v0.22.0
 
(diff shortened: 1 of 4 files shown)


trask (approved):

lmolkova:
Needs #231 to fix build error.

AlexanderWert (approved):

jsuereth (approved):

Oberon00 (commented):
This SHOULD might be redundant because at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation we state;

> The API documentation MUST state that adding attributes at span creation is preferred to calling SetAttribute later, as samplers can only consider information already present during span creation.

arminru:
> This SHOULD might be redundant because at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation we state;
> 
> > The API documentation MUST state that adding attributes at span creation is preferred to calling SetAttribute later, as samplers can only consider information already present during span creation.

@Oberon00 These SHOULD statements still provide an explicit hint that those attributes might be particularly relevant for sampling decisions, which I think is worthwhile in addition to the generic preference stated in the API documentation.

arminru (approved):

Oberon00 (approved):

PR #231, opened 2023-10-31
Update tools to fix CI

lmolkova:


semantic-conventions/src/opentelemetry/semconv/model/utils.py
```diff
@@ -13,7 +13,7 @@
 #   limitations under the License.
 
 import re
-from typing import Tuple
+from typing import Tuple  # noqa: F401
```
lmolkova, line 16: used implicitly, but flake does not understand it

trask (approved):

trask:
@open-telemetry/technical-committee @open-telemetry/specs-semconv-maintainers can you review / merge this to fix the build (and avoid this being a blocker for #230)? thx

AlexanderWert (approved):

semantic-conventions/src/opentelemetry/semconv/model/utils.py
Oberon00: I think the "proper" solution would be to switch from legacy `#type:` comments to using actual annotations.

Oberon00 (approved):

Oberon00:
We could also pin the python minor version when calling the setup-python action to avoid being blocked by such things (although it is nice to update the tools once in a while).

semantic-conventions/src/opentelemetry/semconv/model/utils.py
arminru: I created https://github.com/open-telemetry/build-tools/issues/234 as a follow-up but will merge this PR as-is for now.
cc @lmolkova

arminru (approved):

Issue #253, opened 2024-01-05
Add the semconvgen Library to PyPi (semantic-conventions package)


Kevin-Ngo:
- There are some cases where using the SemanticConvention model outside of a CLI would be useful
- Third-party scripts that perform validation on Semantics on top of the standard syntax validation found in the semconvgen models
- Adding the package to PyPi makes it easily available for other applications/services to reference it as a dependency

Oberon00:
Are we really sure we need this? We have enough compatibility problems and under-documentation with what Jinja templates can access. If we expose this, even more changes will be breaking changes.

At the moment, the issue claims that there are "some cases" where it would be useful. Please list some concrete use cases.

Kevin-Ngo:
The cases that I can think of, and one that I want to use, is wanting access to the syntax validation that is present in the `SemanticConvention` model for the sake of making my own code generation script based off semantic conventions. I don't really need to use the CLI, so having access to the model via a library would make it so that users can use the model (SemanticConvention) and its out of the box syntax validation on the YAML files.

lmolkova:
closing this issue - we're sunsetting otel/semconvgen in favor of [weaver](https://github.com/open-telemetry/weaver) https://github.com/open-telemetry/build-tools/issues/322 and won't make any improvements here.

PR #255, opened 2024-01-08
Update semantic-conventions Directory to Support PyPi Packaging

Kevin-Ngo:
## Description
There are some cases where using the SemanticConvention model outside of a CLI would be useful, such as third-party scripts that perform validation on Semantics on top of the standard syntax validation found in the semconvgen models. That said, adding the package to PyPi makes it easily available for other applications/services to reference it as a dependency. In this PR, a description and package name is edited to allow easy publishing to PyPi.

## Issue
https://github.com/open-telemetry/build-tools/issues/253

## Testing
- [x] Tested publishing the library on PyPi's test server
- [x] Tested running pip install within a venv to install the library

## Steps to Publish
1. Obtain information needed from opentelemetry's PyPi account (API Token) [[more information](https://packaging.python.org/en/latest/tutorials/packaging-projects/#next-steps)]
2. Go to the `semantic-conventions` directory
```bash
cd semantic-conventions
```
3. Install dependencies
```bash
python -m pip install --upgrade pip
pip install --upgrade setuptools wheel twine
```
4. Generate the wheels (distribution) (Note when they are asking for username/password, it should be as follows: 
"**\_\_token\_\_"/{OPENTELEMETRY_ACCOUNT_API_TOKEN}**)
```bash
pip wheel --no-deps . --wheel-dir ./dist
```

5. Upload the distribution
```bash
python3 -m twine upload --repository testpypi dist/* --verbose
```

### DIFF
--- a/semantic-conventions/setup.cfg
+++ b/semantic-conventions/setup.cfg
@@ -13,15 +13,15 @@
 # limitations under the License.
 #
 [metadata]
-name = semconvgen
-description = OpenTelemetry Semantic Conventions utility
+name = opentelemetry-semconvgen
+description = OpenTelemetry Semantic Conventions Utilities
 author = The OpenTelemetry Authors
 author_email = cncf-opentelemetry-contributors@lists.cncf.io
-url = https://github.com/dynatrace-oss-contrib/build-tools/
+url = https://github.com/open-telemetry/build-tools
 platforms = any
 license = Apache-2.0
 classifiers =
-    Development Status :: 1 - Alpha
+    Development Status :: 3 - Alpha
     Intended Audience :: Developers
     License :: OSI Approved :: Apache Software License
     Programming Language :: Python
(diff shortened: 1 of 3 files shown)


joaopgrassi:
We will probably need some sort of OTel credentials to publish this package to PyPi. CC @open-telemetry/technical-committee

Oberon00:
Are we really sure we need this? We have enough compatibility problems and under-documentation with what Jinja templates can access. If we expose this, even more changes will be breaking changes.

semantic-conventions/setup.cfg
```diff
@@ -36,10 +36,10 @@ package_dir=
     =src
 packages=find_namespace:
 install_requires =
-    dataclasses~=0.6
+    dataclasses~=0.6;python_version<'3.7'
```
Oberon00, line 39: I don't think we support python < 3.7 at all, probably we need the Python version we currently run with in the container as minimum.

PR #260, opened 2024-02-06
Sort attribute tables by requirement level and attribute name

pyohannes:
Fixes open-telemetry/semantic-conventions#614

When consulting semantic conventions and the documentation generated from it, especially for metrics, I often find myself asking which attributes I can rely on to be present (which ones are "required"). For me, that's a much more common use case than looking for a single attribute by name.

This changes the ordering of generated attribute tables to order by requirement level: required attributes first, then conditionally required ones, recommended ones, and opt-ins. Especially for metrics, looking for required attributes is a more common use case than looking attributes up by name.

### Example

That's how this would look for `http.server.request.duration` attributes:

|---|---|---|---|---|
| [`http.request.method`](../attributes-registry/http.md) | string | HTTP request method. [1] | `GET`; `POST`; `HEAD` | Required |
| [`url.scheme`](../attributes-registry/url.md) | string | The [URI scheme](https://www.rfc-editor.org/rfc/rfc3986#section-3.1) component identifying the used protocol. [2] | `http`; `https` | Required |
| [`error.type`](../attributes-registry/error.md) | string | Describes a class of error the operation ended with. [3] | `timeout`; `java.net.UnknownHostException`; `server_certificate_invalid`; `500` | Conditionally Required: If request has ended with an error. |
| [`http.response.status_code`](../attributes-registry/http.md) | int | [HTTP response status code](https://tools.ietf.org/html/rfc7231#section-6). | `200` | Conditionally Required: If and only if one was received/sent. |
| [`http.route`](../attributes-registry/http.md) | string | The matched route, that is, the path template in the format used by the respective server framework. [4] | `/users/:userID?`; `{controller}/{action}/{id?}` | Conditionally Required: If and only if it's available |
| [`network.protocol.name`](../attributes-registry/network.md) | string | [OSI application layer](https://osi-model.com/application-layer/) or non-OSI equivalent. [5] | `http`; `spdy` | Conditionally Required: [6] |
| [`network.protocol.version`](../attributes-registry/network.md) | string | Version of the protocol specified in `network.protocol.name`. [7] | `1.0`; `1.1`; `2`; `3` | Recommended |
| [`server.address`](../attributes-registry/server.md) | string | Name of the local HTTP server that received the request. [8] | `example.com`; `10.1.2.80`; `/tmp/my.sock` | Opt-In |
| [`server.port`](../attributes-registry/server.md) | int | Port of the local HTTP server that received the request. [9] | `80`; `8080`; `443` | Opt-In |


semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
```diff
@@ -131,7 +132,12 @@ def _get_attributes(self, templates: Optional[bool]):
                 if templates is None
                 or templates == AttributeType.is_template_type(attr.attr_type)
             ],
-            key=lambda attr: attr.fqn,
+            key=lambda attr: (
+                attr.requirement_level.value
+                if attr.requirement_level
+                else RequirementLevel.OPT_IN.value,
```
lmolkova, line 138: I believe `recommended` is a default requirement level

lmolkova (approved):
LGTM!

(please add a changelog entry)

semantic-conventions/src/opentelemetry/semconv/model/semantic_convention.py
joaopgrassi: Why python is so weird 😆. What about writing this condition in another function? I'm not good at python and this took me a while to figure it out what it is doing :D. Can ignore if this is the python way of doing it. 

```
if attr.requirement_level:
    return attr.requirement_level.value, attr.fqn
return RequirementLevel.RECOMMENDED.value, attr.fqn
```

pyohannes: 1. Switched the default requirement level to `RECOMMENDED`.
2. Extracted the logic for determining the comparison key to a separate function.

joaopgrassi (approved):

lmolkova:
There was a bunch of new merge conflicts - I resolved them. Keeping it open a bit  longer in case anyone wants to take a look.

pyohannes:
> There was a bunch of new merge conflicts - I resolved them. Keeping it open a bit longer in case anyone wants to take a look.

@lmolkova Looks good to me.

Issue #242, opened 2023-11-28
Need a soonish fix for current usage of code generation


AlexanderWert:
- In the past the way build-tools were used for code generation (in many language SDKs) assumed a strict separation between resource attributes and semantic attributes
- This assumption is not true anymore, as there are attributes that can be used as both, resource and semantic attributes
- Currently, there's no (easy/at all?) way to use the build-tools for code generation in a way that the above separation applies if we have namespaces in the registry that would contain both, resource and semantic attributes
- With the migration to a registry-based attributes model, the current limitation 

The above leads to the fact that some work on semantic conventions is blocked, because we cannot introduce namespaces with a mix of resource and semantic attributes.

AlexanderWert:
\cc @lmolkova FYI: I created this issue to track all ongoing semconv PRs that are blocked by this issue.

joaopgrassi:
Given https://github.com/open-telemetry/build-tools/pull/243, do we still need anything else to close this? @AlexanderWert

lmolkova:
Adding a list of remaining things not yet addressed in #243

To merge the PR to main:
- [ ] File names should follow per-language pattern (`client_attrubutes.py`, `ClientAttrubutes.java`) 
- [ ] Update PRs in Java (https://github.com/open-telemetry/semantic-conventions-java/pull/41, https://github.com/open-telemetry/opentelemetry-python/pull/3586) and get approvals from Java and Python SIGs

Can be done as separate PRs:
- [ ] Add examples for attributes and metric generation (common case, split by stability)
- [ ] Improve docs 

Done, needs approvals:
- [ ] Make stability and deprecation independent properties #244

Needs discussion:
- [ ]  Codegen: proper casing for multiword attribute/metric names #599 - AFAIK it's not blocking. None of the affected attributes are stable **yet**

jsuereth:
I created a prototype to help with filenames: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/internal/resourcemapping/resourcemapping.go#L143

TL;DR; is we can allow `--output` filename to be a JINJA2 template itself and provide different versions of the "prefix" for the file to the template,e .g. camel-case, snake-case, ....

lmolkova:
I believe we have just two items left, but I don't consider them blocking:
[...]

PR #263, opened 2024-02-14
Allow --output to be templatized when generating multiple files.

jsuereth:
Related to #242 - "File names should follow per-language pattern"

When outputting multiple files, we allow `--output` to be evaluated as its own jinja template.   This template would have the following values:

- `prefix`
- `pascal_prefix`
- `camel_prefix`
- `snake_prefix`

This should allow language authors to customize filenames as they need for language conventions.


lmolkova (approved):
LGTM!

joaopgrassi (approved):

PR #267, opened 2024-02-19
Adds required stability property to enum members

lmolkova:
See https://github.com/open-telemetry/semantic-conventions/issues/670#issuecomment-1951539146 for the context

As a part of DB and messaging semconv stability, we want to stabilize corresponding `*.system` attribute. 
However it would affect only a subset of corresponding member values.

We need a mechanism that would allow to declare which of the constants are stable.


semantic-conventions/syntax.md
```diff
@@ -377,6 +377,7 @@ An enum entry has the following fields:
 - `value`, string, int, or boolean; value of the enum entry.
 - `brief`, optional string, brief description of the enum entry value. It defaults to the value of `id`.
 - `note`, optional string, longer description. It defaults to an empty string.
+- `stability`, optional stability level, default to `experimental`. Attributes marked as experimental cannot have stable memvers.
```
trask, line 380:
```suggestion
- `stability`, optional stability level, default to `experimental`. Attributes marked as experimental cannot have stable members.
```

trask (approved):

alanwest (approved):

semantic-conventions/src/tests/data/markdown/stability/labels_expected.md
```diff
@@ -7,4 +7,12 @@
...
 | [`test.exp_attr`](labels_expected.md) | boolean |  |  | Required |
 | [`test.stable_attr`](labels_expected.md) | boolean |  |  | Required |
+| [`test.stable_enum_attr`](labels_expected.md) | string |  | `one` | Recommended |
+
+`test.stable_enum_attr` MUST be one of the following:
+
+| Value  | Description |
+|---|---|
+| `one` | member one |
+| `two` | member two |
```
joaopgrassi, line 17: I wonder if it will be clear that the ones without a badge are experimental. I guess that's not a question for enums, but for all attributes we have today.

semantic-conventions/src/tests/data/yaml/errors/stability/experimental_attr_stable_member.yaml
```diff
@@ -0,0 +1,13 @@
+groups:
+  - id: test
+    type: attribute_group
+    brief: 'test'
+    attributes:
+      - id: enum_attr
```
joaopgrassi, line 6: Nit: Even though it's the default, I think it's nice to explicitly add the stability attribute here, so it's very clear what the test/file is about.

joaopgrassi (approved):

semantic-conventions/src/tests/data/yaml/errors/stability/experimental_attr_stable_member.yaml
trask: Given the problems we've had with experimental things being treated as "defacto stable", I think I'd support requiring `stability` attribute on everything

semantic-conventions/src/tests/data/markdown/stability/labels_expected.md
lmolkova: We can start rendering experimental badges in addition to stable ones and include them everywhere, not just in the registry. Created issue on it https://github.com/open-telemetry/semantic-conventions/issues/763.
there is not much to do in this PR though - this is controlled by command line options set in the semconv repo.

semantic-conventions/src/tests/data/yaml/errors/stability/experimental_attr_stable_member.yaml
lmolkova: fixed the test!

Agree that requiring stability to be set explicitly is good - reflected it in this issue https://github.com/open-telemetry/semantic-conventions/issues/763

semantic-conventions/src/tests/data/markdown/stability/labels_expected.md
lmolkova: I think I can still add badge rendering to this PR and also make stability for  enums members required, so unresolving it

lmolkova: updated tests to render experimental badges here, also added deprecated members

semantic-conventions/src/tests/data/yaml/errors/stability/experimental_attr_stable_member.yaml
lmolkova: updated this PR to require stability on enum members in the last commit.

arminru (approved):

PR #271, opened 2024-02-20
Add backward compatibility check mode

lmolkova:
Implements back-compat checks against released version
See https://github.com/open-telemetry/semantic-conventions/issues/740 for the context

Usage: 

`-f path/to/semantic-conventions/model compatibility --previous-version 1.24.0` 

If I run it against current semconv main, I get
```
Found 5 compatibility issues:
                 attribute 'system.processes.status': was removed
                 metric 'system.processes.count': was removed
                 metric 'system.processes.created': was removed
                 attribute 'container.labels': was removed
                 attribute 'k8s.pod.labels': was removed
```


semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -258,9 +307,30 @@ def setup_parser():
...
+    filename = f"v{version}.zip"
+    tmppath = tempfile.mkdtemp()
+    path_to_zip = os.path.join(tmppath, filename)
+    path_to_semconv = os.path.join(tmppath, f"v{version}")
+
+    semconv_vprev = (
+        f"https://github.com/open-telemetry/semantic-conventions/archive/{filename}"
+    )
+
+    request = requests.get(semconv_vprev, allow_redirects=True, timeout=30)
```
trisch-me, line 325: shouldn't we check 200? if unknown version is provided, 404 with html instead of zip file will be returned

semantic-conventions/src/opentelemetry/semconv/main.py
```diff
@@ -258,9 +307,30 @@ def setup_parser():
...
 
 
+def download_previous_version(version: str) -> str:
+    filename = f"v{version}.zip"
+    tmppath = tempfile.mkdtemp()
+    path_to_zip = os.path.join(tmppath, filename)
+    path_to_semconv = os.path.join(tmppath, f"v{version}")
+
+    semconv_vprev = (
+        f"https://github.com/open-telemetry/semantic-conventions/archive/{filename}"
```
trisch-me, line 322: is it correct url? doesn't work for me.
https://github.com/open-telemetry/semantic-conventions/archive/refs/tags/v1.24.0.zip seems to working

semantic-conventions/src/opentelemetry/semconv/templating/compatibility.py
```diff
@@ -0,0 +1,191 @@
...
+            if cur.stability != prev.stability:
+                problems.append(
+                    Problem(
+                        "attribute",
+                        prev.fqn,
+                        f"stability changed from '{prev.stability}' to '{cur.stability}'",
+                    )
+                )
+
+            if isinstance(prev.attr_type, EnumAttributeType):
```
jsuereth, line 80: Should we prevent *any* type change on a stable attribute?

jsuereth (approved):
This is a great additiion!  Thanks for putting this together.

- I echo Alexandra's concerns around download success.  However, I don't feel the need to reapprove on fix.
- I had one other comment on additioinal stability checking we can do, but consider it non-blocking.

semantic-conventions/src/opentelemetry/semconv/templating/compatibility.py
lmolkova: we do (line 102), it just needs to happen in different ways for enums and non-enums

semantic-conventions/src/opentelemetry/semconv/main.py
lmolkova: good catch!

semantic-conventions/src/tests/data/compat/enum_value_removed/vprev.yaml
```diff
@@ -0,0 +1,17 @@
...
+    brief: "first group"
+    note: "first group note"
+    prefix: "first"
+    attributes:
+      - id: third_attr
+        type:
+          members:
+            - id: enum_one
+              brief: "enum one"
+              value: "one"
```
joaopgrassi, line 13: This example is the same as `enum_type_changed`. Perhaps we want to in `vnext` rename the attribute entirely so it really shows it was removed?

semantic-conventions/src/tests/data/markdown/enum_extend_list/expected.md
```diff
@@ -0,0 +1,24 @@
...
+<!-- semconv second -->
+| Attribute  | Type | Description  | Examples  | Requirement Level |
+|---|---|---|---|---|
+| `first.enum_attr` | string |  | `"one"` | Recommended |
+
+`first.enum_attr` MUST be one of the following:
+
+| Value  | Description |
+|---|---|
+| `one` | one |
```
joaopgrassi, line 22: Sorry maybe it's silly but where does `one` and `two` are coming from here? 🤔 I don't really get the `enum_extend_list` tests here. Seem they are also not used in any test?

joaopgrassi (commented):
This is great! Thanks for working on it. 

Just a question I had while reviewing/reading the issue: Who is the target group of this new task in the tool? Like, when is it going to be executed? During a release of semconv? By SDK authors? Ran automatically as part of a CI/CD in semconv? It's not clear and maybe this should be stated somewhere?

semantic-conventions/src/opentelemetry/semconv/main.py
trisch-me: Thanks @lmolkova, works for me today, direct url as well, probably was some glitch

semantic-conventions/src/tests/data/compat/enum_value_removed/vprev.yaml
lmolkova: `enum_type_changed` tests `"one"` -> `1`, i.e. string to int
This check is redundant (since we check that value remains the same), but I'd still prefer to keep it for the sake of better error message.

But your comment made me realize that I mistakenly allowed member id to change (as long as value remains) - this should not be allowed since codegen would do `enum Foo {  {id} = {value} }` and both should remain intact.

I renamed this test case to `value_changed` and added one more for `member_removed`

Thanks!

semantic-conventions/src/tests/data/markdown/enum_extend_list/expected.md
lmolkova: another great catch, removed them

lmolkova:
@joaopgrassi my intention is to add it to CI checks - here's draft PR that demonstrates it https://github.com/open-telemetry/semantic-conventions/pull/764

joaopgrassi (approved):

PR #327, opened 2024-06-27
Simplify dependabot config

lucacome:
Dependabot now supports wildcard and we can simplify the config by making it search in all the directories automatically.


lmolkova:
Thank you for the contribution @lucacome !

Do you know if there is a way to test it?

lucacome:
@lmolkova the only way I know is to merge this PR in your own fork.

arminru (approved):

Issue #240, opened 2023-11-24
Add `embed` concept for attributes reuse (in addition to `ref`)


AlexanderWert:
This issue describes a concrete proposal on [how to realise attributes reuse](https://github.com/open-telemetry/semantic-conventions/issues/339) through yaml syntax and tooling.

## Related issues / PRs:
- https://github.com/open-telemetry/semantic-conventions/pull/351
- https://github.com/open-telemetry/semantic-conventions/pull/559

## Proposal

I'm proposing to introduce a new, additional type of attribute reference called `embed` (alternatively: `nest`) in the syntax that would behave similarly to `ref` property BUT with the following difference: When `embed` is used the referenced attribute will be nested under the `namespace` of the attribute group in which the attribute is being referenced. 

## Example
See the following example that illustrates the idea and the difference to `ref`:

We would define geo attributes in a `geo.yaml` file:

```
groups:
  - id: geo
    prefix: geo
    type: attribute_group
    brief: Describes the geo location
    attributes:
      - id: location.lon
        type: double
        brief: Longitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
        examples: [ -73.614830 ]
      - id: location.lat
        type: double
        brief: Latitude of the geo location in [WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84).
        examples: [ 45.505918 ]
```

In client and server files we would `embed` the attributes.
! Note: Here I consciously `embed` the  `geo.location.lon` attribute BUT `ref` the `geo.location.lat` attribute to demonstrate the difference:

```
groups:
  - id: client
    prefix: client
    type: attribute_group
    brief: ...
    attributes:
      - id: address
        ...
      - embed: geo.location.lon            <<-- THAT IS NEW
      - ref: geo.location.lat
```

When rendering (or code generating) the above it will result in the following attributes:

- `client.geo.location.lon`
- `geo.location.lat`

Note the difference that `geo.location.lon` is now nested under the `client` namespace because it was embedded (`embed`) while `geo.location.lat` keeps it's original full name because it was referenced (`ref`).

## Extended proposal

We could even think about an `embed_namespaces` property on the `attribute_group` level to simplify embeddings of entire namespaces into other namespace. For example for `geo` we could do something like the following to embed ALL of the `geo.*` fields into the client:

```
groups:
  - id: client
    prefix: client
    type: attribute_group
    brief: ...
    embed_namespaces:
[...]

PR #332, opened 2024-07-03
Support for embedding of single attributes

trisch-me:
update build tools to support embedding of attribute implemented in [weaver](https://github.com/open-telemetry/weaver/pull/212)


trisch-me:
Partial fix for https://github.com/open-telemetry/build-tools/issues/240

semantic-conventions/src/tests/data/markdown/ref_embed/server.yaml
```diff
@@ -0,0 +1,15 @@
...
+    brief: "Describes information about the server."
+    attributes:
+      - id: id
+        type: string
+        stability: experimental
+        brief: >
+          some server id
+        examples: ['123']
+      - ref: geo.lat
+        prefix: true
```
lmolkova, line 14: I believe we discussed `- embed: geo.lat`  instead of `- ref: geo.lat` and `prefix: true`.

What would be the reason to not do `embed: geo.lat` based on the original design and do the `prefix` flag instead?

Why I'd prefer a new verb (`embed`)
- `ref` always means the same thing - reference attribute and keep the original name.
- obvious when reading. Imagine 10 other lines between `ref` and `prefix` and how hard it'd be to understand what name is supposed to be generated
- easier to check in policies
- `prefix` feels like a string property and not a boolean flag.

trisch-me: we have discussed it few weeks ago while I have started to implement it in weaver, so that `ref` and `embed` are the same in functionality and difference is only in prefixing the parent namespace or not. Hence this approach was selected. Embed will stay for the whole namespace. 

I have announced this change also in embed issue linked in this PR

AlexanderWert (approved):

semantic-conventions/src/tests/data/markdown/ref_embed/server.yaml
lmolkova: it still has all the problems listed above, so I'd like to revisit this choice

trisch-me: I can address these problems because IMO they are not critical
- ref means reference and this is also referencing attribute from another namespace. The difference is in prefix of the attribute. So `embed` means `ref` + `prefix` of the parent group. And this is how it was implemented
- I personally don't feel this way, we do have a lot of things, like for example ref itself can override description etc. They all means something therefore they are written there
- It's additional check for policy (`ref` + `prefix`) instead just (`embed`), but I don't feel it's something complicated
- regarding string property - this is probably due to the fact that we already have such keyword for the group. 

The main reason why we went this way was as described before complete similarity between ref and new embed, therefore we have decided to go with it. 

I would like to get more opinions on this @jsuereth @lquerel

lmolkova: So you agree that these are problems, critical or not and ideally we would have one-line that expresses that attribute is embedded instead of two?

It'd also be confusing to have `ref` to embed one attribute and `embed` to embed all. 

If we want to support both, let's find a good option with `embed_one + `embed_all`, `embed` + `embed_all`, etc. 

I'm struggling to come up with a real-life use-case for "embed-all-attributes from" and I see a lot of places where embed-one would work great. So both of the above options seem reasonable to me.

lmolkova: another option could be to do 

```yaml
- embed: foo.bar # for attribute name
- embed: foo.* # for the namespace
```

but it seems a bit subtle.

lmolkova: > The main reason why we went this way was as described before complete similarity between ref and new embed, therefore we have decided to go with it.

Similarity in what sense? On the tooling implementation side? `Ref` and `embed` are different on the schema and semconv definition level.

trisch-me: we do have already an option to embed all and embed only one. They are written in the original issue. Embeded namespaces will be defined on the top level of the group, while embedded attribute is defined under attributes. 

> I'm struggling to come up with a real-life use-case for "embed-all-attributes from" 

we do have a lot of such real examples. The most prominent one is about `geo` namespace included into `client` or `server`. Then `process` included into `process` as parent node and so on.

> Similarity in what sense?

In both. On tolling side it has the same implementation with additional prefix. On semantical meaning they also both are references to another namespace with difference in prefix.

Let's discuss it tomorrow during our tooling meeting

lmolkova: Let's get back to the original concern: having prefix: true under reference has problems (critical or not). We should try to design schema to be easy to use, easy to read, etc. The "not a critical problem" is not a good argument. 

Having `embed_one` and `embed_all` verbs would result in much better outcome for schema users. So why the pushback?

trisch-me: As I have stated before, this decision was made due to similarity of ref and embed functionality. I don't see this as a problem otherwise I wouldn't go and implement it. 
Therefore I propose to discuss it once more during our meeting

lmolkova: I do see above problems with this approach, which would be resolved if we used `embed_one` or similar with no visible downsides. 

The embedding is semantically similar to definition of the attribute (as it creates new attribute).

lmolkova (changes requested):
Let's reconsider reusing `ref`. Suggestion is to use `embed_one`, more details in discussion

semantic-conventions/src/tests/data/markdown/ref_embed/server.yaml
trisch-me: We have discussed it yesterday and I will propose new ideas after a meeting with @lquerel today
Thank you for raising your concerns here.

PR #493, opened 2026-07-20
Major updates to build-tools Protobuf docker image

jsuereth:
- Use commit SHA instead of raw checkouts
- Verify tars, zips, etc. against known SHA
- Update to latest, compatible, versions (caveat some patch files to fix compatibility issues)
- Update README to denote this new policy/goal
- Remove dependabot - use Rennovate, so this has a chance of staying up-to-date (or we detect when new versions are released)


.github/renovate.json5
```diff
@@ -0,0 +1,40 @@
...
+    'helpers:pinGitHubActionDigestsToSemver',
+  ],
+  packageRules: [
+    {
+      groupName: 'all patch and minor versions',
+      matchUpdateTypes: [
+        'patch',
+        'minor',
+      ],
+      schedule: [
```
florianl, line 18: Make sure we don't get duplicates from the second rule:
```suggestion
      ],
      excludeManagers: [
        'github-actions',
        'dockerfile',
      ],
      schedule: [
```

protobuf/Dockerfile
```diff
@@ -1,26 +1,57 @@
...
+ARG UPX_VERSION=5.2.0
+ARG UPX_AMD64_SHA256=3db5d3294707439db97866feab8d75d800f028f48481a40547411824da4288a1
+ARG UPX_ARM64_SHA256=55d48a61e8ffd17152db871c855376cba7f08e830b37799d0947a16dff8ec36c
 
 ARG TARGETARCH=amd64
 
-FROM alpine:3.18 as protoc_builder
-RUN apk add --no-cache build-base curl automake autoconf libtool git zlib-dev linux-headers cmake ninja
+FROM alpine:3.21@sha256:48b0309ca019d89d40f670aa1bc06e426dc0931948452e8491e3d65087abc07d AS protoc_builder
+RUN apk add --no-cache build-base curl automake autoconf libtool git zlib-dev linux-headers cmake ninja pkgconfig pkgconf abseil-cpp-dev patch
```
florianl: `pkgconfig` is a symlink package to `pkgconf`
```suggestion
RUN apk add --no-cache build-base curl automake autoconf libtool git zlib-dev linux-headers cmake ninja pkgconf abseil-cpp-dev patch
```

jsuereth: Fixed

.github/renovate.json5
jsuereth: Fixed

protobuf/README.md
```diff
@@ -88,3 +88,11 @@ Use this command to find the path to proto files included in the image:
 ```bash
 docker run --rm -it --entrypoint=/bin/sh otel/build-protobuf:latest -c "find /usr/include -name *.proto"
 ```
+
+## SLSA Compliance & Maintenance Requirements
```
pellared, line 92: (non-blocking) I do not think this is required for SLSA Compliance, but I think that we should also sign the artifacts using `cosign`.

Reference: https://github.com/open-telemetry/sig-security/blob/main/docs/recommendations.md#sign-release-artifacts-with-sigstore-cosign

This can be addressed in a follow-up PR.

protobuf/README.md
```diff
@@ -88,3 +88,11 @@ Use this command to find the path to proto files included in the image:
...
+
+To ensure Supply chain Levels for Software Artifacts (SLSA) compliance and build reproducibility:
```
pellared, line 94: Shouldn't we document how users can verify (attest)? I mean adding exact end-user instructions.

protobuf/README.md
jsuereth: Good point - yes.  I'll add that as a follow-up task.

protobuf/README.md
jsuereth: Great point - I'll add that

protobuf/README.md
jsuereth: https://github.com/open-telemetry/build-tools/issues/494

protobuf/README.md
jsuereth: Added!

protobuf/Dockerfile
```diff
@@ -31,62 +62,93 @@ RUN git clone --recursive --depth=1 -b v${GRPC_VERSION} https://github.com/grpc/
...
 ARG PROTOBUF_C_VERSION
+ARG PROTOBUF_C_SHA256
 RUN mkdir -p /protobuf-c && \
-    curl -sSL https://api.github.com/repos/protobuf-c/protobuf-c/tarball/v${PROTOBUF_C_VERSION} | tar xz --strip 1 -C /protobuf-c && \
+    curl -sSL https://github.com/protobuf-c/protobuf-c/releases/download/v${PROTOBUF_C_VERSION}/protobuf-c-${PROTOBUF_C_VERSION}.tar.gz -o /tmp/protobuf-c.tar.gz && \
+    echo "${PROTOBUF_C_SHA256}  /tmp/protobuf-c.tar.gz" | sha256sum -c - && \
+    tar xzf /tmp/protobuf-c.tar.gz --strip 1 -C /protobuf-c && \
+    rm /tmp/protobuf-c.tar.gz && \
     cd /protobuf-c && \
+    patch -p1 < /tmp/protobuf-c-pr797.patch && \
```
jack-berg, line 84: These look like manual patches applied to the contents of the release. I don't understand what's motivating this or what it does.

jsuereth: We're building on alpine linux, and using the Alpine linux (MUSL) version of abseil to build these libraries.  I'm not sure why - but basically C++ is failing to understand certain `absl::string_view` and `std::string` implicit conversions - so these patches basically remove the ambiguity in those type coersions so things compile.

I'm not happy about having them - but it was necessary to pull along to the latest version *if* we continue to build from source here.

I could try to look into pre-built binaries - but I was trying to start "small" and just bump versions here initially.

jack-berg: I'm fine with this but think we should have an explanatory comment.

Happy to approve and defer to you, since despite the TC being the maintainers, we clearly have not acted in a maintainer capacity in some time.

jack-berg (approved):

.github/workflows/protobuf-dockerimage.yml
```diff
@@ -23,25 +25,38 @@ jobs:
...
+    - name: Build and Push Docker image with Attestations
+      uses: docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4 # v6.15.0
+      with:
+        context: ./protobuf
+        platforms: linux/${{ matrix.TARGETARCH }}
+        push: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+        load: ${{ !(github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')) }}
+        tags: ${{ steps.meta.outputs.tags }}
+        provenance: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
+        sbom: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
```
pellared: nit: add EOL

```suggestion
        sbom: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}

```

protobuf/README.md
```diff
@@ -88,3 +88,27 @@ Use this command to find the path to proto files included in the image:
...
+
+```bash
+# Verify build provenance against the OpenTelemetry organization
+gh attestation verify oci://docker.io/otel/build-protobuf:latest \
+  --owner open-telemetry
+
+# Inspect the bundled Software Bill of Materials (SBOM) attestation
+gh attestation inspect oci://docker.io/otel/build-protobuf:latest \
+  --owner open-telemetry
+```
```
pellared: It would be more secure  to use more precise args (--repo, --signer-workflow, --source-ref) like proposed in https://github.com/open-telemetry/sig-security/pull/300

.github/renovate.json5
```diff
@@ -0,0 +1,44 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+  extends: [
+    'config:best-practices',
+    'helpers:pinGitHubActionDigestsToSemver',
+  ],
+  packageRules: [
+    {
+      groupName: 'all patch and minor versions',
```
pellared, line 9: Why not bumping major version of GitHub Actions?

protobuf/README.md
pellared: Btw have you tested it? :sweat_smile:

pellared: Should it not be something like the following?

```
image=docker.io/otel/build-protobuf@sha256:IMAGE_DIGEST
docker buildx imagetools inspect "${image}"  --format '{{ json .Provenance.SLSA }}'
```

florianl (approved):

lmolkova (approved):

protobuf/README.md
jsuereth: I haven't tested it because we can't verify attestations until we publish with them. I have tested build attestations on weaver.

Would you be ok if I drop this README for now and add it in *after* we have attestations and I can verify which syntax works?

.github/renovate.json5
jsuereth: I'm not a a rennovate expert - this is copy-pasta from another repo.  Should we bump major versions?  I can make that change

jsuereth: Looking elsewhere - I think this is due to compatibility assumptions.  Bumping major versions MAY break and should be a separate PR issued.  The reason for this bundling is because minor/patch updates are likely to merge easily, while major are more likely to break, so we don't bundle them together.

Line 23 is where major version PRs are created - IIUC

protobuf/Dockerfile
jsuereth: I added the explanatory comment - should hopefully help keep this up to date in the future.

I'm probably going to add an `AGENT.MD` next so you can trick co-pilot into fixing version bumps by having it understand what these patches do, in case we run into issues with renovate bumping versions.

protobuf/README.md
jsuereth: Removed it completely for now - Post release we'll update the docs - opened an issue to track.

pellared: > I haven't tested it because we can't verify attestations until we publish with them. 

You can test it on a fork.

.github/renovate.json5
pellared: > Line 23 is where major version PRs are created - IIUC

👍 

BTW major versions of GitHub Actions are published often when the action moving to a newer Node.js runtime or a newer minimum version of the self-hosted is required.

pellared (approved):

protobuf/README.md
jsuereth: My account doesn't have any room left for pushing artifacts (builds have been failing for me until I go clean up some personal projects) - so no I can't test on my own fork.

jsuereth: Ok - I added the attestations for `gh` as well - so we'll have BOTH `gh` and docker attestations with this flow now.   I added a comment on your PR @pellared with what I believe should be working here.

